PR #23, opened 2026-06-01
fix(dhcp): re-apply KEA config when running config drifts back to boo…

dmathur0:
…tstrap

Ported from kiwi-platform MR !240. When the kea container is recycled it reloads the baked-in bootstrap config (no remote lease-database), but the sync loop only re-applied on Redis change, so the pod deadlocked with /healthcheck returning 500. Adds _kea_running_config_diverged() to detect the drift and force a re-apply, plus a looser backstop livenessProbe on config-sync-v4 so kubelet can recover if the in-loop check itself wedges.

## Validation

- [ ] Standard CI passes.
- [ ] Kind integration passes, or this PR explains why it was not run. https://github.com/NVIDIA/nv-config-manager/actions/runs/28128905479

Passing Kind Integration run:

## Checklist

- [ ] Commits are signed off for DCO compliance.

## Summary by CodeRabbit

* **Bug Fixes**
  * DHCP KEA sync now detects post-restart configuration divergence for remote (PostgreSQL) lease databases and re-applies the correct lease-db settings, even when the stored config is unchanged.

* **Improvements**
  * Added a conservative “backstop” liveness probe to the config-sync service to trigger restarts when needed.
  * Healthcheck logic and lease-database injection now consistently follow the configured remote vs local mode.
  * The sync loop performs a best-effort running-state probe and forces resync when KEA falls back to bootstrap behavior.

* **Tests**
  * Added unit and async end-to-end tests covering divergence detection, resync behavior, local-mode skipping, and query error handling.


src/nv_config_manager/dhcp/cli.py
```diff
@@ -208,6 +208,44 @@ def refresh_kea_configuration(
     asyncio.run(_refresh_loop_async(ip_version, check, refresh_interval))
 
 
+async def _kea_running_config_diverged(
```
zsblevins: I would rename this, diverged implies you're doing a direct comparison, which would be the cleanest approach but would always report divergence as kea has a lot of defaults and minor formatting changes between what we generate and what it reports back as its config. In the absence of being able to do a true divergence test, we should name this in a way that indicates it's only checking for the default configuration state

dmathur0: Good call. Renamed it to `_kea_running_config_is_bootstrap` (and updated the call site/tests/docstring) to make clear it only detects the bootstrap-default state rather than performing a true divergence comparison. Pushed in dfc6342.

src/nv_config_manager/dhcp/cli.py
```diff
@@ -218,6 +256,7 @@ async def _sync_kea_configuration_async(
     ini_config = load_config()
     kea_client = KeaClient.from_config(ini_config, attached=True)
     redis_client = RedisClient.from_config(ini_config)
+    remote_lease_db = not ini_config["dhcp.lease_db"].getboolean("local")
```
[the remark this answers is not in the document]
dmathur0: Addressed in dfc6342. Added a shared `is_remote_lease_db()` helper in `common/config.py` that reads `[dhcp.lease_db] local` via `getboolean(..., fallback=None)` and logs a warning + defaults to remote mode when the section/option is missing. Now used in `cli.py`, `api.py`, and `kea_dhcp_confgen.py` so startup/healthcheck no longer crash on an incomplete INI.

dmathur0:
Split the Helm-only backstop `livenessProbe` (the `deploy/helm/templates/dhcp.yaml` change) out into #65 so it can ship independently now. This PR's KEA bootstrap-drift detection / re-apply **application logic** stays here and should be held until after QA. Once #65 merges I'll drop the dhcp.yaml hunk from this branch so the two don't overlap.

PR #24, opened 2026-06-01
feat(nautobot): YAML-driven group-mapping RBAC for JWT-authenticated users

dmathur0:
## Summary

YAML-driven, IdP-group → Nautobot `ObjectPermission` RBAC for JWT-authenticated users. This is the **feature half** of the original combined PR; the integration test suite and CI wiring were split out into the stacked follow-up **#122** so this can be reviewed/merged on its own.

- **`components/nautobot/nv_config_manager_auth/rbac.py`** — reads a YAML group-mapping file and reconciles Django `Group` memberships + managed `<group>_<action>` `ObjectPermission` records per login. Fail-closed, three-state config semantics, automatic revocation when entries/actions are removed. Group lifecycle is operator-managed by default (`autoCreateGroups` opt-in).
- **`jwt_authentication.py`** — wraps the RBAC sync in `web_request_context(user)` so change-logging signals have a user to attribute (fixes an `AttributeError` crash during permission deletes).
- **Helm** — `groupMapping` three-state model (unconfigured / configured-empty / configured), guarded on explicit key presence so the default deploy never revokes-all.
- **Docs** — new "Authentication and RBAC" section in `docs/nautobot/nautobot-integration.mdx`.
- **Unit tests** — `test_rbac.py` + additions to `test_jwt_authentication.py` cover mapping parsing, permission/membership sync, revocation, superuser reconciliation, and the change-context regression.

Integration/live-cluster coverage lives in **#122** (stacked on this branch).

## Test plan

- [x] Unit tests: `uv run pytest components/nautobot/tests/`
- [x] Local e2e via `make kind-up-sec` + `scripts/rbac-local-test/verify.sh` across all three group-mapping states

## Summary by CodeRabbit

* **New Features**
  * Added opt-in YAML-driven JWT RBAC synchronization for Nautobot, mapping IdP groups to Django Groups and per-action permissions, including `is_superuser: true` handling via an inert membership marker.
  * Reconciliation now runs on every login with improved superuser promotion/demotion controls, including deterministic behavior when the mapping is configured but empty or partially unavailable.

* **Helm / Deployment**
  * Chart rendering and pod rollout checksum are now conditional on `groupMapping`; ConfigMap mounting changed from single-file to directory.
  * Added chart support for `autoCreateGroups`.

* **Documentation**
  * Updated Authentication and RBAC guidance for mapping states, revocation behavior, and authentication-time reconciliation.

* **Tests**
  * Expanded coverage for YAML parsing/validation, fail-closed sync, and login-context-scoped reconciliation.

## Kind Integration

Kind integration completed with **success** for [`fff3ebce2a0c`](https://github.com/NVIDIA/nv-config-manager/actions/runs/29443770561).


deploy/helm/templates/_nautobot-configmap-data.tpl
```diff
@@ -76,6 +76,18 @@ NV_CONFIG_MANAGER_JWT_COOKIE: {{ .Values.oidc.cookieName | default "NVConfigMana
...
+# login.  Rendered by the chart into the nautobot group-mapping ConfigMap and
+# mounted at the path below.
+NV_CONFIG_MANAGER_GROUP_MAPPING_PATH: "/app/config/group-mapping.yaml"
+{{- if .Values.nautobot.rbac.autoCreateGroups }}
+# Opt-in: nv_config_manager_auth.rbac will create Django Groups referenced in the
+# mapping on the fly the first time a logging-in user matches them.  Default off --
+# operators are normally expected to create the Group rows up-front.
+NV_CONFIG_MANAGER_AUTO_CREATE_GROUPS: "true"
+{{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
dmathur0: Fixed in e39912a. The group-mapping ConfigMap, volume, volumeMount, checksum annotation, and the `NV_CONFIG_MANAGER_GROUP_MAPPING_PATH` env var are now all guarded on `{{- if hasKey .Values.nautobot.rbac "groupMapping" }}`. With the key absent, nothing renders and `rbac.mapping_is_configured()` sees the feature as unconfigured, so no sync/revocation runs.

Verified with `helm template` in all three states:
- default (key omitted) → 0 group-mapping resources
- `groupMapping: []` → ConfigMap with `groups: []` + volume + mount + env var (deliberate revoke-all)
- `groupMapping: [ ... ]` → full ConfigMap content

deploy/helm/values.yaml
```diff
@@ -2381,6 +2381,86 @@ nautobot:
...
+    #           content_types: ["all"]
+    #         change:
+    #           content_types: ["ipam.*"]
+    #         delete:
+    #           content_types: ["ipam.ipaddress"]
+    #           constraints:
+    #             status__name: "Active"
+    #     - name: "nv-config-manager-admins"
+    #       is_superuser: true
+    groupMapping: []
```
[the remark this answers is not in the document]
dmathur0: Fixed in e39912a. The default is now the key **commented out** (omitted) rather than `[]`, so the out-of-the-box behavior is "unconfigured" and existing `is_superuser`/group memberships are left untouched — consistent with the `superuserGroups: []` semantics.

Semantics are now keyed on presence (`hasKey`), not truthiness:
- key omitted (default) → unconfigured, no ConfigMap/mount, no sync
- `groupMapping: []` → configured-but-empty, the explicit revoke-everyone idiom
- `groupMapping: [ ... ]` → configured

Documented the three states inline in `values.yaml` and in a new "Authentication and RBAC" section of the Nautobot integration docs.

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,654 @@
...
+
+log = logging.getLogger(__name__)
+
+
+DEFAULT_MAPPING_PATH = "/app/config/group-mapping.yaml"
+ALL_CONTENT_TYPES = "all"
+_TRUTHY = frozenset({"1", "true", "yes", "on"})
+
+# Django/Nautobot internal models that should NOT be touched by an "all"
+# expansion.  Cerebro maintains a similar list -- granting view/change on
```
zsblevins: Don't reference Cerebro in public code (it'll get a rename on OSS)

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,654 @@
...
+    """Load and normalise the group-mapping YAML.
+
+    Returns a ``{group_name: group_config}`` dict.  An empty dict is returned
+    when the file does not exist, is empty, or contains no mappings -- the
+    feature is opt-in and should never crash a login when unconfigured.
+    """
+    path = path or _mapping_path()
+
+    # Funnel every filesystem failure mode into ``GroupMappingError`` so the
+    # caller's documented fail-closed path always runs.  The previous
```
zsblevins: Clean up some of the comments here, referencing "previous patterns" that never existed in actual released code is a bit confusing

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,654 @@
...
+    # from "non-mapping top-level" (``false`` / ``0`` / ``[]`` / ``""`` --
+    # operator error, must surface).  The previous ``yaml.safe_load(fh) or {}``
+    # collapsed every falsy value into ``{}`` and bypassed the validator
+    # below, silently disabling RBAC.
+    if data is None:
+        return {}
+    if not isinstance(data, dict):
+        raise GroupMappingParseError(f"{path}: top-level must be a mapping, got {type(data).__name__}")
+
+    # Accept "groups" (native) or "azure_app_roles" (cerebro-compat).  We
```
zsblevins: Another cerebro reference

zsblevins: I'd also remove the azure references altogether and the azure fallback, there's nothing we need to fallback to on a new implementation

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,654 @@
...
+) -> None:
+    """Reconcile *user*'s Django Groups and ObjectPermissions against *mapping*.
+
+    The caller is responsible for the "is the feature configured?" gate
+    via :func:`mapping_is_configured` -- this function deliberately does
+    NOT short-circuit on an empty mapping.  When the operator opts into
+    the feature and then writes ``groupMapping: []`` (or the YAML fails
+    to parse), pass 1 + pass 2 become no-ops and pass 3 prunes everything
+    this module previously granted.  Conflating "empty mapping" with
+    "feature disabled" is how previously-granted memberships /
```
zsblevins: More references to non-existent "previous" implementations in this context

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,731 @@
...
+
+        content_types = _resolve_content_types(action_config.get("content_types") or [])
+        constraints = action_config.get("constraints") or {}
+        if not isinstance(constraints, dict):
+            log.warning(
+                "rbac: group %r action %r constraints must be a mapping; coercing to {}",
+                group.name,
+                action,
+            )
+            constraints = {}
```
ryanheffernan: If constraints is in the wrong format, the code silently removes the restrictions but still grants the permission, so a typo could give someone broader access than intended. It would be safer to reject or skip that permission instead.

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,731 @@
...
+
+        perm, _created = ObjectPermission.objects.update_or_create(
```
ryanheffernan, line 796: Permission names are shared across all of Nautobot. If another group or user already uses the generated name, this code will reuse and overwrite their permission, potentially changing their access. The same issue applies to the membership marker above. We should either verify ownership before updating the permission or fail when there is a collision

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,731 @@
...
+                    "rbac: Django Group %r is in the mapping but does not exist in Nautobot; "
+                    "create it (admin UI / fixture / Job) or set NV_CONFIG_MANAGER_AUTO_CREATE_GROUPS=true",
+                    name,
+                )
+                continue
+
+        # ``is_superuser`` groups don't get per-action ObjectPermissions --
+        # the user's global superuser flag obviates them.
+        if config.get("is_superuser"):
+            _prune_group_object_permissions(group)
```
ryanheffernan: This creates an edge case:
- user x logs in with a superuser group defined in the YAML, joins the django admins group
- the group mapping is removed from the YAML
- user x logs in again, they are no longer a superuser but remain in the Django admins group, and retain any manual permissions that group had

If this is a concern then we may need to keep a record of this former membership after prune, which it appears is done by the `_apply_group_permission_config(group, {})` call below

ryanheffernan (changes requested):

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,781 @@
...
+                "rbac: group %r action %r config must be a mapping, got %s; skipping",
+                group.name,
+                action,
+                type(action_config).__name__,
+            )
+            continue
+        perm_name = f"{group.name}_{action}"
+        # Keep the name off the prune list in every branch below (create, skip,
+        # or collision) so we never delete a row we deliberately left untouched.
+        kept_perm_names.add(perm_name)
```
ryanheffernan: Since we mark this permission as kept before validating constraints, a malformed value is skipped while an existing permission remains. Nautobot supports both a dictionary and a list of dictionaries here, so we should validate both forms before marking the permission as kept. Invalid constraints should leave it unkept so the existing managed permission is revoked.

```
constraints:
  - vid__gte: 100
    vid__lt: 200
  - status__name: Reserved
```
and then do something like this to safely validate and revoke
```
perm_name = f"{group.name}_{action}"

if _object_permission_owned_by_other(perm_name, group):
    kept_perm_names.add(perm_name)
    log.warning(...)
    continue

constraints = action_config.get("constraints", {})

valid_constraints = isinstance(constraints, dict) or (
    isinstance(constraints, list)
    and bool(constraints)
    and all(isinstance(item, dict) and bool(item) for item in constraints)
)

if not valid_constraints:
    log.warning(...)
    continue  # Existing managed permission is not kept, so it is pruned.

kept_perm_names.add(perm_name)
```

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,781 @@
...
+        * ``"ipam.foo.bar"`` / ``".prefix"`` / ``"ipam."``
+        * non-strings.
+
+    Unknown / not-installed models (valid shape, no such row) are also
+    skipped with a warning rather than raising, so a single typo doesn't
+    block a login.
+    """
+    if not spec:
+        return []
+    if ALL_CONTENT_TYPES in spec:
```
ryanheffernan, line 465: spec is typehinted to a list[str] but the type passed in from `action_config` is ANY, so this could be a str. if that string is e.g. 'wall' then 'all' in 'Wall' will return true and it will grant the action across all types. we should validate that spec is a list before this check

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,781 @@
...
+                action,
+                type(action_config).__name__,
+            )
+            continue
+        perm_name = f"{group.name}_{action}"
+        # Keep the name off the prune list in every branch below (create, skip,
+        # or collision) so we never delete a row we deliberately left untouched.
+        kept_perm_names.add(perm_name)
+
+        if _object_permission_owned_by_other(perm_name, group):
```
ryanheffernan, line 765: we check ownership here but then do the update on line 751 below, which creates a potential for a race if another transaction modifies the permissions bewteen these two points. it's porabbly worth trying to perform these two operations within one atomic transaction

components/nautobot/nv_config_manager_auth/rbac.py
daveshm: good find

components/nautobot/nv_config_manager_auth/rbac.py
daveshm: https://github.com/NVIDIA/nv-config-manager/pull/24/changes/BASE..9e4008f6f86200eb3bc6b7a3f5d9cc47f0d850d2#diff-fb7163705262b86bc5fe7d74057265045a4a9387b02579aabaf82d49426711adR467 the entire call chain is decorated as atomic

daveshm: i think this would also require two transactions to be touching the same thing in the same millisecond or so

ryanheffernan: just doing some reading on how this works - `@transaction.atomic` just guarantees commit/rollback of this transaction, it does not protect against another transaction occuring at the same time as this one since we don't lock the DB when we do the ownership check - only when we do the update. Another transaction can modify the table between the ownership check and the update.

You're right though that this is a pretty narrow window for a race, so I think it's fair to accept this to minimize complexity in an already large MR - perhaps a followup hardening task could be created to audit and harden the RBAC schema?

components/nautobot/nv_config_manager_auth/rbac.py
```diff
@@ -0,0 +1,826 @@
...
+            )
+            continue
+
+        # Validate constraints BEFORE marking the perm as kept.  Nautobot accepts
+        # a mapping or a non-empty list of mappings; anything else is malformed.
+        # A malformed value is skipped WITHOUT keeping the name, so an existing
+        # managed row is pruned below (fail closed: revoke rather than grant an
+        # unconstrained perm or silently retain the last-good one).
+        constraints = action_config.get("constraints")
+        if constraints is None:
```
ryanheffernan, line 782: just want to confirm I understand the desired behavior here - is this accurate? 

```
constraints: []
```
No permissions granted

```
constraints: {}
```
All permissions granted

```
constraints: None
```
All permissions granted

deploy/helm/values.yaml
```diff
@@ -2442,6 +2442,106 @@ nautobot:
...
+    #
+    # Revocation: removing a whole entry (``- name: foo``) revokes access
+    # automatically on each affected user's next login -- they are removed
+    # from the Django Group and its ``<group>_<action>`` ObjectPermissions
+    # are detached and (when no other group references them) deleted.  The
+    # Django Group row itself is left in place so operators retain final
+    # say over the group catalog; purely-manual groups (no managed
+    # ``<group>_<action>`` perms attached) are never touched.
+    #
+    # Limitation: removing *the last* ``is_superuser: true`` entry from the
```
ryanheffernan: nit: update this to match the new behavior

deploy/helm/values.yaml
```diff
@@ -2442,6 +2442,106 @@ nautobot:
...
+    #                   uses, via JwtProviderConfig.claim_groups).
+    #   is_superuser:   optional bool; when true, the user is promoted to
+    #                   Django superuser as long as they hold this group.
+    #                   Per-action permissions are then irrelevant.
+    #   nautobot_permissions:
+    #     view / add / change / delete / run:
+    #       content_types: list of "app_label.model" (or the shortcuts
+    #                      "all" and "app_label.*").  Unknown models are
+    #                      logged and skipped, not raised.
+    #       constraints:   optional dict of Django-ORM lookups applied to
```
ryanheffernan: same nit as above

components/nautobot/nv_config_manager_auth/rbac.py
daveshm: `constraints: []` is when the constraints are malformed, this will actually remove the entire permission as well
`constraints: {}` unconstrained 
`constraints: None` same as `{}`

components/nautobot/nv_config_manager_auth/rbac.py
dmathur0: Tracked as a follow-up hardening task: #131 (close the TOCTOU window between the ownership check and `update_or_create` via create-then-recheck on `IntegrityError` or a name-scoped advisory lock). Accepting the narrow residual here to keep this MR scoped.

ryanheffernan (approved):

PR #40, opened 2026-06-10
Add configurable resource values for temporal archive service

spidercensus:
## Description

We need to add a way to specify resources for the archive container in the values file the same way that we do for the other temporal services in this template.

Without this specification the defaults wiil apply, and in some environments this will lead archive service to be OOM killed.

## Validation

- [ ] Standard CI passes.

Passing Kind Integration run:
https://github.com/NVIDIA/nv-config-manager/actions/runs/27309824019

## Checklist

## Release Notes

* **Chores**
  * Updated Helm deployment so the Temporal archive component correctly receives defined CPU and memory requests/limits.
  * Fixed template rendering so the resource settings apply to the archive container as intended, ensuring more reliable scheduling and resource enforcement during deployment.


spidercensus:
Waiting on kind integration test here: 
https://github.com/NVIDIA/nv-config-manager/actions/runs/27308231581

jojung1 (approved):

PR #65, opened 2026-06-25
fix(helm): add backstop livenessProbe to DHCP config-sync sidecar

dmathur0:
## Description

Split out of #23 so this low-risk, **Helm-only** change can ship independently while the KEA bootstrap-drift detection/re-apply **application logic in #23 waits for QA**.

Adds a conservative backstop `livenessProbe` to the `config-sync-v4` container in `deploy/helm/templates/dhcp.yaml`. If the in-loop KEA divergence check ever wedges (asyncio loop stalls, or the kea control socket stays unresponsive long enough that `get_config()` keeps raising), kubelet recycles the sidecar. On restart the container re-runs the unconditional "Run once" `set_config()` and heals the deadlock — even if the in-loop check itself is broken.

- Probes the healthcheck endpoint (`/healthcheck` on port `9090`) in the same pod/network namespace, matching the pattern used by the kea container's probe.
- Thresholds are deliberately looser than the other probes (`failureThreshold * periodSeconds = 300s` of failure) so the in-loop recovery (~10s) wins under normal conditions and we only recycle the sidecar when the loop itself is the problem.

This contains **no application code** — it's purely the Kubernetes safety net, so it carries minimal risk and does not need to wait on QA of the drift-recovery logic.

## Validation

- [x] `helm template test . --values values-ci.yaml` renders cleanly with the new probe on the config-sync container.
- [ ] Standard CI passes.
https://github.com/NVIDIA/nv-config-manager/actions/runs/28194950934/job/83519142880
## Checklist

- [x] New or existing tests cover these changes, or the PR explains why tests are not needed. (Helm-only probe addition; covered by Kind integration rendering.)
- [ ] Generated artifacts are updated when applicable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an additional health check to improve recovery for DHCP config synchronization.
  * Helps the service restart more reliably if the sync process becomes unresponsive.

### DIFF
--- a/deploy/helm/templates/dhcp.yaml
+++ b/deploy/helm/templates/dhcp.yaml
@@ -182,6 +182,30 @@ spec:
         - name: NV_CONFIG_MANAGER_INI
           value: {{ include "nv-config-manager.vaultAgent.configManagerIniPathDhcp" . | quote }}
         {{- include "nv-config-manager.customLabelsEnv" . | nindent 8 }}
+        # Backstop livenessProbe for the case where the in-loop divergence
+        # check in _sync_kea_configuration_async can't recover (e.g. the
+        # asyncio loop wedges or the kea control socket is unresponsive long
+        # enough that get_config() keeps raising). On failure, kubelet
+        # recycles this container, which restarts the Python process and
+        # runs the unconditional "Run once" set_config() at startup --
+        # heals the deadlock even if the in-loop check is broken.
+        #
+        # Probes the healthcheck sidecar's endpoint (same pod, same network
+        # namespace, so kubelet reaches port 9090 regardless of which
+        # container serves it -- same pattern as the kea container's probe).
+        #
+        # Thresholds are deliberately looser than the kea/healthcheck/api
+        # probes (failureThreshold * periodSeconds = 300s of failure) so the
+        # in-loop check (~10s recovery) wins under normal conditions and we
+        # only recycle the sidecar if the loop itself is the problem.
+        livenessProbe:
+          httpGet:
+            path: /healthcheck
+            port: 9090
+          initialDelaySeconds: 180
+          periodSeconds: 30
+          timeoutSeconds: 5
+          failureThreshold: 10
         resources:
           {{- toYaml .Values.networkDhcp.resources.confgen | nindent 10 }}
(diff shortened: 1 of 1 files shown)


deploy/helm/templates/dhcp.yaml
```diff
@@ -182,6 +182,30 @@ spec:
...
+        # namespace, so kubelet reaches port 9090 regardless of which
+        # container serves it -- same pattern as the kea container's probe).
+        #
+        # Thresholds are deliberately looser than the kea/healthcheck/api
+        # probes (failureThreshold * periodSeconds = 300s of failure) so the
+        # in-loop check (~10s recovery) wins under normal conditions and we
+        # only recycle the sidecar if the loop itself is the problem.
+        livenessProbe:
+          httpGet:
+            path: /healthcheck
```
ryanheffernan, line 203: If the goal of this is to detect when the config-sync job is stuck, I don't think it will work because /healthcheck reports on the [general DHCP service liveness state](https://github.com/NVIDIA/nv-config-manager/blob/26d5ee88320dd30c51aaf2d8b937b8fad6c2dd09/src/nv_config_manager/dhcp/api.py#L141-L142), not the state of the config-sync job. 

If we do want to detect and recover from the config sync job getting stuck we could do something like set an overall timeout on the asyncio event loop, or add some kind of heartbeat mechanism where the sync loop writes a timestamp to a file and a liveness probe defined in helm checks it for progress. Example:

```
livenessProbe:
  exec:
    command:
    - python
    - -c
    - |
      import pathlib, time, sys
      p = pathlib.Path("/tmp/config-sync-v4.heartbeat")
      sys.exit(0 if p.exists() and time.time() - float(p.read_text()) < 300 else 1)
  initialDelaySeconds: 180
  periodSeconds: 30
  timeoutSeconds: 5
  failureThreshold: 3
```

ryanheffernan: Discussed further and this is specifically to address the scenario where DHCP reverts to a bootstrap config (https://github.com/NVIDIA/nv-config-manager/pull/23), which it will detect since healthchecks fail in that scenario.

ryanheffernan (approved):

PR #69, opened 2026-06-26
fix(installer): Detect early cloud-init completion

ryanheffernan:
## Description

Fix false Air simulation setup failures caused by two issues in the cloud-init watcher:

1. The remote setup can finish before the local watcher observes the completion banner in its live `tail -f` stream. When `cloud-init status` becomes `done`, the watcher previously treated the missing streamed banner as failure even when `/var/log/nvcm-setup.log` showed that setup completed successfully.
2. Stopping the SSH log-tail process can exceed the five-second grace period. The resulting `subprocess.TimeoutExpired` exception then masks the setup result and reports the overall run as failed.

This change:

- checks both `/var/log/cloud-init-output.log` and `/var/log/nvcm-setup.log` for an existing completion marker before starting the live tail;
- rechecks the remote logs after cloud-init reports `done` before declaring setup failure;
- moves the completion banner to the end of the setup script, after cluster status and next-step output, so it represents full script completion;
- centralizes log-tail cleanup, escalating from `terminate()` to `kill()` when the process does not exit within five seconds;
- prevents cleanup timeouts from overwriting an already-determined success or failure result.

## Validation

- Kind integration was not run because this only changes the Air installer watcher and its child-process cleanup; no cluster manifests or runtime components changed.

## Checklist

- [x] New or existing tests cover these changes.
- [x] No documentation update is required for this false-failure correction.
- [x] No generated artifacts are affected.


ryanheffernan:
/ok to test [77c0f4b](https://github.com/NVIDIA/nv-config-manager/pull/69/commits/77c0f4b4060ba8a70cba82951f0ee140b9cca119)

PR #89, opened 2026-07-01
fix: address Sonar reliability and security findings

ryanheffernan:
## Description

Address high/med sonar findings and address issue preventing the sonar pipeline.

- Scope GitHub Actions permissions to the jobs that require them and pass the RC source ref through the step environment instead of interpolating it into shell code.
- Make shell function returns explicit and replace legacy single-bracket Bash conditionals with `[[ ... ]]`.
- Propagate image-target resolution failures through every CI caller and preserve Helm/CTR failures in the air-gapped bundle paths.
- Normalize multi-package Nautobot coverage XML to one source root with package-qualified filenames so Sonar can resolve every file unambiguously.
- Share the Nautobot coverage commands between merge-request and main jobs through one fail-fast YAML anchor.
- Normalize a trailing slash in `SONAR_HOST_URL` before calling the SAST export API so the job receives JSON instead of the Sonar HTML application.
- Disable the unused UFM mock service-account token, reserve `128Mi` of ephemeral storage, and cap it at `512Mi`.
- Mark the Linked VRF attribute labels as semantic table row headers while preserving Nautobot's 25% label-column width.

## Validation

- Parsed every GitHub Actions workflow as YAML and verified affected checkout jobs have explicit `contents` permissions with no workflow-level grants remaining.
- Ran `bash -n` on every modified shell script, exercised the air-gapped creator's `--help` path, and confirmed the affected Bash scripts contain no single-bracket conditionals.
- Exercised `image_target_env.sh` with a representative image-target record and verified its exact exports.
- Verified a missing image target exits the caller with status 1, a synthetic Helm exit 42 reaches the fallback under `set -e`, and a synthetic NVCR exit 7 is retried three times.
- Ran all 124 Nautobot tests with coverage, regenerated the normalized XML, and verified it contains one source root with package-qualified paths while preserving 66% coverage.
- Parsed the shared Nautobot coverage anchor, verified both jobs resolve to the same block, and checked its shell syntax.
- Reproduced the SAST export failure against the double-slash API URL and verified the normalized URL returns a valid GitLab SAST JSON document.
- Parsed the UFM mock manifest and asserted token automounting is disabled and both storage request and limit are present.
- Verified the Linked VRF table contains three scoped row headers.
- Built the Nautobot overlay plugin sdist and wheel with `uv build`.
- Ran `git diff --check`.

- [x] Standard CI passes.
- [x] Kind integration was not run because these changes are static-analysis remediations to CI permissions, build utilities, a development-only mock manifest, and template semantics; no application workflow behavior changed.

Passing Kind Integration run: https://github.com/NVIDIA/nv-config-manager/actions/runs/28533963918

## Checklist

- [x] Existing static checks and targeted validation cover these changes; no new application tests are needed.
- [x] No documentation update is needed because there is no user-facing behavior change.
- [x] Generated artifacts are not affected; generated Go bindings were intentionally left unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation, guard checks, and exit-code handling across air-gapped bundle creation, image packaging, and target-selection logic.
  * Hardened the Kea admin entrypoint to fail fast on missing/invalid config and streamline database validation.
* **Style**
  * Updated the VXLAN retrieve UI table markup for clearer, more accessible label semantics.
* **Chores**
  * Tightened CI token permissions to job scope and improved CI shell execution safety when sourcing image-target environment exports.
  * Refined coverage generation and SonarQube URL handling; updated the UFM mock deployment’s token and storage configuration.

### DIFF
--- a/.github/workflows/create-rc-tag.yml
+++ b/.github/workflows/create-rc-tag.yml
@@ -22,10 +22,6 @@ on:
         default: true
         type: boolean
 
-permissions:
-  actions: read
-  contents: read
-
 concurrency:
   group: release-tagging-${{ github.repository }}
   cancel-in-progress: false
@@ -34,6 +30,9 @@ jobs:
   validate:
     name: Validate RC Tag
     runs-on: linux-amd64-cpu4
+    permissions:
+      actions: read
+      contents: read
     outputs:
       commit: ${{ steps.resolve.outputs.commit }}
       go_tag: ${{ steps.version.outputs.go_tag }}
@@ -167,6 +166,7 @@ jobs:
           COMMIT: ${{ steps.resolve.outputs.commit }}
           GO_TAG: ${{ steps.version.outputs.go_tag }}
           RC_TAG: ${{ steps.version.outputs.rc_tag }}
+          SOURCE_REF: ${{ inputs.source_ref }}
         run: |
           {
             echo "### RC tag request"
@@ -174,7 +174,7 @@ jobs:
             echo "- Tag: \`${RC_TAG}\`"
             echo "- Go module tag: \`${GO_TAG}\`"
             echo "- Commit: \`${COMMIT}\`"
-            echo "- Source ref: \`${{ inputs.source_ref }}\`"
+            echo "- Source ref: \`${SOURCE_REF}\`"
           } >> "$GITHUB_STEP_SUMMARY"
 
   create-tag:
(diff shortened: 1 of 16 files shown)


ryanheffernan:
/ok to test [c84cb3f](https://github.com/NVIDIA/nv-config-manager/pull/89/commits/c84cb3ff2fa0ac0bc5e383d4f6703820a48ed7a6)

zsblevins (approved):

PR #95, opened 2026-07-02
(fix) Update pkey add call - fix for UFM behavior

susanhooks:
## Description

**Problem**
Adding a member with membership: limited reported a workflow failure, but UFM had already added the member as full. The add activity sent the requested memberships to UFM's Add endpoint (POST /resources/pkeys/) under the plural memberships array. On UFM 6.19.1, POST silently ignores memberships and defaults every added GUID to full — so the write "succeeded" (200) with the wrong type, and the downstream verify stage then failed on the mismatch.

**Root cause**
Confirmed empirically against UFM 6.19.1 via curl probes. The Add and Set endpoints are asymmetric:

POST /resources/pkeys/ (Add): honors only a single membership string; ignores the plural memberships array (defaults to full); rejects a list under membership (400).
PUT /resources/pkeys/ (Set): honors the index-aligned plural memberships array.
The client used the plural form on POST, which that endpoint doesn't read.

**Fix**
Reworked add_guids_to_pkey into an atomic read-merge-write:

GET the partition's current members.
Merge in the requested GUIDs (requested membership wins on conflict; existing members preserved, so the op stays additive).
Issue a single PUT of the full set, which UFM applies atomically and honors per-GUID via memberships.

## Validation

Updated mock ufm for real behavior, curls against the UFM verified that the POST to add didn't work, but the PUT did. 
- [x] Standard CI passes.

Passing Kind Integration run:

## Checklist

## Summary by CodeRabbit

- **New Features**
  - PKey member updates now preserve existing memberships when adding GUIDs, rather than replacing the full set.
  - Membership updates now handle both single-value and per-GUID membership inputs more consistently.

- **Bug Fixes**
  - Fixed cases where membership data could be interpreted incorrectly during add and update flows.
  - Improved handling of empty or missing member data so existing partitions are reported more reliably.

- **Tests**
  - Updated coverage for add/update behavior, merged memberships, and validation of membership payloads.

### DIFF
--- a/src/nv_config_manager/temporal/ngc/activities/ib_pkey.py
+++ b/src/nv_config_manager/temporal/ngc/activities/ib_pkey.py
@@ -88,7 +88,8 @@ class AddGuidsInput(BaseModel):
     """Parameters for adding port GUIDs to an existing PKey partition.
 
     ``memberships`` is index-aligned with ``guids`` (one "full"/"limited" per
-    GUID), the per-port form UFM accepts via the ``memberships`` array.
+    GUID). The activity merges these into the partition's current members and
+    issues a single PUT, since UFM's Add endpoint cannot set per-GUID membership.
     """
 
     host: str
@@ -111,7 +112,8 @@ class SetGuidsInput(BaseModel):
     """Parameters for atomically setting a PKey's exact GUID membership.
 
     ``memberships`` is index-aligned with ``guids`` (one "full"/"limited" per
-    GUID), the per-port form UFM accepts via the ``memberships`` array.
+    GUID), the per-port form UFM's Set endpoint (PUT) accepts via the
+    ``memberships`` array.
     """
 
     host: str
@@ -163,6 +165,36 @@ def _validate_memberships_aligned(pkey: str, guids: list[str], memberships: list
             f"guids length ({len(guids)})",
             non_retryable=True,
         )
+
+
+async def _get_pkey_state(client: UFMClient, pkey: str) -> tuple[bool, dict[str, str], bool | None]:
+    """Read a PKey's current members from UFM."""
+    try:
+        pkey_data = await client.request(
+            "GET",
+            f"/resources/pkeys/{pkey}",
+            params={"guids_data": "true"},
+        )
+    except UFMClientError as e:
+        if e.status_code != 404:
+            raise
+        return False, {}, None
+
+    if not isinstance(pkey_data, dict):
+        raise ApplicationError(
+            f"PKey {pkey} returned an unexpected response from UFM",
+            non_retryable=True,
+        )
+
+    members: dict[str, str] = {}
+    for entry in pkey_data.get("guids", []):
+        if isinstance(entry, dict):
+            members[str(entry.get("guid", "")).lower()] = str(entry.get("membership", "")).lower()
+        else:
+            members[str(entry).lower()] = ""
+
+    ip_over_ib = pkey_data.get("ip_over_ib")
+    return True, members, ip_over_ib if isinstance(ip_over_ib, bool) else None
 
 
 def _parse_pkey_int(pkey_str: str) -> int:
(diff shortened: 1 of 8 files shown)


src/nv_config_manager/temporal/ngc/activities/ib_pkey.py
```diff
@@ -298,7 +330,13 @@ async def verify_pkey_created(input: VerifyPKeyInput) -> VerifyPKeyOutput:
...
 @activity.defn
 async def add_guids_to_pkey(input: AddGuidsInput) -> AddGuidsOutput:
-    """Add port GUIDs to an existing PKey partition on UFM."""
+    """Add port GUIDs to an existing PKey partition on UFM.
+
+    UFM's Add endpoint (POST) cannot assign per-GUID membership in one call and
+    silently defaults added members to "full". To assign membership reliably and
+    without leaving partial state, this reads the partition's current members, merges
+    in the requested GUIDs, and issues a single PUT that UFM applies atomically.
+    """
```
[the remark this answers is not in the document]
susanhooks: Will not address, it's covered via temporal checks later in the workflow.

src/nv_config_manager/temporal/ngc/activities/ib_pkey.py
```diff
@@ -309,25 +351,35 @@ async def add_guids_to_pkey(input: AddGuidsInput) -> AddGuidsOutput:
...
-        len(input.guids),
-        input.pkey,
-        input.host,
-        input.guids,
-        input.memberships,
-    )
-
     async with UFMClient(host=input.host, site=input.site) as client:
-        await client.request("POST", "/resources/pkeys/", json=payload)
+        _, current_members, current_ip_over_ib = await _get_pkey_state(client, input.pkey)
```
ryanheffernan, line 355: since this operation isn't atomic, concurrent workflow runs can overwriting or undoing each other's changes - this might need some kind of [mutex](https://github.com/temporalio/samples-go/tree/main/mutex)  or use POSTs instead of PUTs to append to the membership lists rather than replace them

src/nv_config_manager/temporal/ngc/activities/ib_pkey.py
```diff
@@ -309,25 +351,35 @@ async def add_guids_to_pkey(input: AddGuidsInput) -> AddGuidsOutput:
...
+            merged[guid.lower()] = membership.lower()
+
+        payload: dict[str, Any] = {
+            "pkey": input.pkey,
+            "guids": list(merged.keys()),
+            "memberships": list(merged.values()),
+            "ip_over_ib": (
+                current_ip_over_ib if current_ip_over_ib is not None else input.ip_over_ib
+            ),
+            "index0": input.index0,
```
ryanheffernan, line 368: should this value be preserved from UFM rather than overwritten if it isn't provided in the input ?

src/nv_config_manager/temporal/ngc/activities/ib_pkey.py
susanhooks: coderabbit raised the same issue, but there's two things 1) this is explicitly a workaround for the UFM API not doing what it says - the POST call does not actually accept mixed membership. When you try to append via post, it defaults all of them to the same membership regardless of what is passed in, that's what this PR is trying to address. 
2) This is an add path that only adds, and then it verifies later in the workflow with a fresh read that would fail and retry via temporal if expected =/= actual. 

I think the mutex path could be added to be extra careful, but would prefer to add that later as a follow up.

ryanheffernan (approved):

src/nv_config_manager/temporal/ngc/activities/ib_pkey.py
susanhooks: it's essentially dead code, I can strip it out if needed. index0 is always false, as that's the management pkey created on boot

PR #96, opened 2026-07-02
Feat/kiwi otel all services

susanhooks:
## Validation

- [x] Standard CI passes.

Passing Kind Integration run:

https://github.com/NVIDIA/nv-config-manager/actions/runs/29451433161
## Checklist

## Summary
* **New Features**
  * Added centralized OpenTelemetry tracing bootstrap and automatic tracing for Config Store, DHCP, MCP, Render, ZTP, and Temporal.
  * Updated Helm observability to use a single top-level `observability` toggle with shared `OTEL_*` injection and tracing opt-out annotations.
* **Bug Fixes**
  * Startup continues even if optional outbound HTTP client instrumentation fails.
* **Tests**
  * Added automated coverage for OTLP target selection and idempotent/concurrent tracing initialization, including Temporal telemetry behavior.
* **Documentation**
  * Documented tracing configuration, OTLP endpoint selection, and provided Helm examples.
* **Chores**
  * Renamed the MCP executable entrypoint to `nv-config-manager-mcp`.

Screenshot 2026-07-15 at 3 08 58 PM
Screenshot 2026-07-15 at 3 12 06 PM

## Kind Integration

Kind integration completed with **success** for [`d15264a0974b`](https://github.com/NVIDIA/nv-config-manager/actions/runs/29451433161).


deploy/helm/templates/_helpers.tpl
```diff
@@ -1504,23 +1504,27 @@ Usage: {{ include "nv-config-manager.externalDnsHostname" .Values.networkZtp.ing
...
+config-store/dhcp/ztp/render/mcp and the Temporal worker/api/scheduler/archive).
+Call sites guard the include with `if .Values.observability.enabled`.
 
-temporal.observability.otlpEndpoint is optional. When unset the endpoint
-defaults to the in-cluster Alloy receiver in the release namespace:
-  http://alloy.<namespace>.svc.cluster.local:4317
-Set it explicitly to target a different collector (e.g. the managed cluster
-OTLP collector on ngcops-eks).
+Points the OTel SDK at an existing OTLP collector:
+  - observability.otlpEndpoint when set (managed/external collector, e.g. Panoptes), else
```
zsblevins: avoid referencing panoptes, that's an internal tool

deploy/helm/values.yaml
```diff
@@ -1530,6 +1530,22 @@ networkDhcp:
...
 # =============================================================================
+# Observability — OTel tracing (global toggle)
+# =============================================================================
+# One switch for OpenTelemetry trace export across every HTTP API
+# (config-store, dhcp, ztp, render, mcp) and the Temporal worker/api/scheduler/
+# archive. When enabled, each service's OTel SDK exports OTLP spans (and Temporal
+# runtime metrics) via standard OTEL_* env vars injected by the chart — no
+# per-pod sidecar. Backend-agnostic: the collector fans traces out to whatever
+# it is wired to (local Tempo via values-observability.yaml, Panoptes in prod).
+observability:
```
zsblevins, line 1533: We have other monitoring blocks which will make this location a bit confusing, I'd put this in the monitoring doc with a tracing section instead

deploy/helm/templates/_helpers.tpl
susanhooks: fixed

deploy/helm/templates/config-store.yaml
```diff
@@ -67,6 +67,9 @@ spec:
           value: {{ include "nv-config-manager.vaultAgent.configManagerIniPathApp" . | quote }}
         - name: NV_CONFIG_MANAGER_SERVICE
           value: "config_store.api"
+        {{- if .Values.observability.enabled }}
+        {{- include "nv-config-manager.otelAppEnv" (dict "root" . "serviceName" "nv-config-manager-config-store") | nindent 8 }}
+        {{- end }}
```
ryanheffernan, line 75: the `nv-config-manager.temporalServerOtelOptOutAnnotations` helper is included in `temporal.yaml` for the temporal pod but not the other pods, so if we want to be able to opt out the other services we'll need to include something like that in these pods too

deploy/helm/values.yaml
susanhooks: moved to monitoring docs

deploy/helm/templates/temporal.yaml
```diff
@@ -430,8 +430,8 @@ spec:
         - name: BMC_CREDS_PATH
           value: "/vault/secrets/bmc-creds.json"
         {{- end }}
-        {{- if .Values.temporal.observability.enabled }}
-        {{- include "nv-config-manager.temporal.otelAppEnv" (dict "root" . "serviceName" "nv-config-manager-temporal-worker") | nindent 8 }}
+        {{- if .Values.observability.enabled }}
+        {{- include "nv-config-manager.otelAppEnv" (dict "root" . "serviceName" "nv-config-manager-temporal-worker") | nindent 8 }}
```
ryanheffernan, line 434: just a note: helm doesn't error on unknown keys, so if we upgrade any environment that has `temporal.observability.enabled: true` without setting `observability.enabled` then tracing will silently stop for that deployment. i doubt this is an issue for us since we just added this today and haven't deployed it, and argo should handle setting the right values. if we wanted to guard against silent failures like this we could add values json schemas but that's way outside the scope of this :)

deploy/helm/templates/config-store.yaml
susanhooks: good catch, added

docs/overview/observability.mdx
```diff
@@ -62,6 +62,48 @@ Key service metrics include:
...
+
+Instrumented components:
+
+| Component | `OTEL_SERVICE_NAME` |
+| :-------- | :-------------------- |
+| Config Store API | `nv-config-manager-config-store` |
+| DHCP API | `nv-config-manager-dhcp` |
+| ZTP API | `nv-config-manager-ztp` |
+| Render API | `nv-config-manager-render` |
+| MCP server | `nvidia-config-manager-mcp` |
```
ryanheffernan: nit: should we make this consistent with the other service names? `nv-config-manager-mcp`

deploy/helm/templates/temporal.yaml
susanhooks: understood, thanks

docs/overview/observability.mdx
susanhooks: I think so - this matches though the name [here](https://github.com/NVIDIA/nv-config-manager/blob/main/src/nv_config_manager/mcp/main.py#L64). I think normalizing all references makes sense?

susanhooks: I renamed all the pieces in the last commit.

ryanheffernan (approved):

PR #105, opened 2026-07-07
add workflow lock mechanism

susanhooks:
## Validation

https://github.com/NVIDIA/nv-config-manager/actions/runs/28898508569
- [x] Standard CI passes.

Passing Kind Integration run:

## Checklist

## Summary
* **New Features**
  * Added Redis-backed distributed async locking with a safe local fallback, including token-based acquire/renew/release behavior.
  * Enabled metadata-driven Temporal workflow locking with automatic TTL renewal and unlock on workflow failure.
  * Added per-(host, pkey) concurrency control for IB PKey member workflows, including host canonicalization.
* **Refactor**
  * Centralized shared Temporal lock activities and ensured they’re registered across Temporal workers.
  * Dynamic workflow endpoints now canonicalize request input before starting workflows.
* **Tests**
  * Expanded unit/integration coverage for Redis locking and end-to-end Temporal lock behavior and canonicalization.


src/nv_config_manager/common/lock.py
```diff
@@ -0,0 +1,181 @@
...
+from nv_config_manager.common.config import is_local_environment, redis_client
+
+if TYPE_CHECKING:
+    from types import TracebackType
+
+    from nv_config_manager.common.client import RedisClient
+
+log = logging.getLogger(__name__)
+
+log = logging.getLogger(__name__)
```
ryanheffernan: dup lines

src/nv_config_manager/temporal/common/activities/lock.py
```diff
@@ -0,0 +1,90 @@
...
+
+class ReleaseWorkflowLockInput(BaseModel):
+    """Parameters for releasing a held workflow lock."""
+
+    key: str
+    token: str
+
+
+@activity.defn
+async def acquire_workflow_lock(input: AcquireWorkflowLockInput) -> None:
```
ryanheffernan, line 56: if a workflow acquires a lock and then fails to record it for some reason, then retrying this activity would likely result in failing to acquire the lock for the remainder of the TTL. not sure how likely/important this is so just noting in case there is an easy fix

src/nv_config_manager/temporal/ngc/workflows/ib_pkey_member_add.py
```diff
@@ -124,6 +125,7 @@ class IBPKeyMemberAddWorkflow(WorkflowMetadataMixin, StageMixin, ArchiveMixin):
     workflow_input_class = IBPKeyMemberAddInput
     workflow_api_endpoint = "/ngc/ib_pkey_member_add"
     workflow_namespace = "ngc"
+    workflow_lock = WorkflowLockSpec(key_fields=["host", "pkey"])
```
ryanheffernan, line 129: host can be either a device name or IP so we need to make sure to account for that so we don't end up allowing multiple locks for the same host (same for the 2 other IB workflows)

src/nv_config_manager/temporal/common/decorators/workflow.py
```diff
@@ -54,3 +88,98 @@ async def _run(*args: object) -> Any:
...
+            fail_on_conflict=fail_on_conflict,
+        ),
+        start_to_close_timeout=timedelta(
+            seconds=spec.wait_timeout_seconds + _LOCK_ACTIVITY_BUFFER_S
+        ),
+        retry_policy=_FAIL_FAST_RETRY_POLICY if fail_on_conflict else _WAIT_RETRY_POLICY,
+    )
+
+
+async def _renew_loop(key: str, token: str, spec: WorkflowLockSpec) -> None:
```
ryanheffernan, line 166: since the redis lock is 120s and starts before this loop starts, it's possible we just miss the deadline and the lock becomes free before we renew it. we may need to adjust the timers to add some buffer

src/nv_config_manager/temporal/common/decorators/workflow.py
```diff
@@ -54,3 +88,98 @@ async def _run(*args: object) -> Any:
...
+async def _acquire_lock(key: str, token: str, spec: WorkflowLockSpec) -> None:
+    """Block until the per-resource lock is held (or fail fast on conflict)."""
+    fail_on_conflict = spec.on_conflict == "fail"
+    await workflow.execute_activity(
+        acquire_workflow_lock,
+        AcquireWorkflowLockInput(
+            key=key,
+            token=token,
+            ttl_seconds=spec.ttl_seconds,
+            wait_timeout_seconds=spec.wait_timeout_seconds,
```
ryanheffernan, line 156: with fail_on_conflict=true we still wait for wait_timeout_seconds before failing, is this intended?

src/nv_config_manager/common/lock.py
susanhooks: fixed

src/nv_config_manager/temporal/ngc/workflows/ib_pkey_member_add.py
susanhooks: added mixin to resolve name/ip before workflow- maybe overkill

src/nv_config_manager/temporal/common/decorators/workflow.py
susanhooks: added buffer

src/nv_config_manager/temporal/common/decorators/workflow.py
susanhooks: no, fixed in recent commit, ty

src/nv_config_manager/temporal/common/activities/lock.py
susanhooks: addressed in commit

ryanheffernan (approved):

PR #115, opened 2026-07-09
support kgateway as an alternate gateway API implementation

zsblevins:
## Description

NMC landed on kgateway as the Gateway API solution instead of Envoy. In order to support multiple pluggable gateway implementations, we're moving the OIDC integration out of the gateway and back to oauth2-proxy.

## Validation

- [ ] Standard CI passes.
- [ ] Kind integration passes, or this PR explains why it was not run.

The kind integration test is manual due to taking ~30 min to complete. When the PR is ready for
review, approve the current commit and start the suite with these PR comments:

```text
/ok to test <sha>
/kind test
```

As a fallback, run Actions -> Kind Integration -> Run workflow against the copy-pr-bot generated
`pull-request/<PR_NUMBER>` branch. Use the default `test_path` for the full suite, or narrow it
only while debugging.

The completed workflow posts its conclusion and exact run URL as a PR comment.

Passing Kind Integration run:

## Checklist

- [ ] I am familiar with the contributing guidelines in `CONTRIBUTING.md`.
- [ ] Commits are signed off for DCO compliance.
- [ ] New or existing tests cover these changes, or the PR explains why tests are not needed.
- [ ] Documentation is updated for user-facing behavior changes.
- [ ] Generated artifacts are updated when applicable, such as OpenAPI specs,
      docs screenshots, or Helm/rendered outputs.

## Summary by CodeRabbit

- **New Features**
  - Added controller-aware **kgateway** support for secured local deployments, including new secured targets and gateway controller selection.
  - Extended **Gateway API + OIDC** Helm/installer behavior to support **Envoy Gateway** and **kgateway**, including Gateway parameters and expanded multi-backend OIDC proxy rendering.
- **Bug Fixes**
  - Improved authentication to accept legacy JWT cookies and **proxy bearer tokens**; tightened logout behavior with safer redirect handling and cookie cleanup.
- **Documentation**
  - Updated SECURITY, installer/TUI, Helm samples, and air-gapped/OCI/pinned image guidance to match the new gateway/proxy conventions and versions.

## Kind Integration

Kind integration completed with **success** for [`c17f1c75636e`](https://github.com/NVIDIA/nv-config-manager/actions/runs/29269631301).

PR #122, opened 2026-07-10
test(nautobot): gated RBAC group-mapping integration suite + CI security toggle

dmathur0:
## Summary

Stacked on top of #24 (base branch = `port/nautobot-rbac-group-mapping-mr238`). This is the **test/CI half** of the original combined RBAC PR, split out per review so the core feature can land independently.

Contains only the test harness and CI wiring — no product/auth code:

- **`src/tests/integration/test_rbac_group_mapping.py`** — live-cluster suite (gated behind `--rbac`) covering: group-mapping ConfigMap presence, group + `ObjectPermission` grants on login, superuser promotion, and a regression for the delete-during-reconcile crash.
- **`src/tests/integration/conftest.py`** — `--rbac`/`--rbac-*` options, `rbac` marker, and fixtures. Tokens are minted headlessly in-process via the OIDC Resource Owner Password grant against the local Keycloak (`rbac_get_token`), so no browser/pre-provisioned token is needed.
- **`.github/workflows/kind-integration.yml`** — opt-in `security` input that deploys the Keycloak + SPIRE stack (`make kind-up-sec`), moves the chart into the group-mapping CONFIGURED state, and scopes the run to the RBAC suite (the rest of the suite is not yet auth-aware under the security deploy).
- **`scripts/rbac-local-test/values-configured.yaml`** — local Helm override that moves a `make kind-up-sec` deploy into the group-mapping CONFIGURED state (the same override the workflow above applies).
- **`pyproject.toml`** — registers the `rbac` marker.

## Follow-up

Converging this with @zoeblevins' headless-token wiring (#115) so `kind-integration` can run the full suite under the security stack (single auth path, `security` on by default) is tracked as follow-up.

## Test plan

- [ ] Merge #24 first; this PR's base retargets to `main` automatically on merge.
- [ ] Dispatch `kind-integration` with `security=true` and confirm the RBAC suite passes.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in **security** mode to the integration workflow, including configured-state rollout and auth-aware test execution.
  * Added live **RBAC group-mapping** end-to-end coverage via a new `rbac` pytest marker and `--rbac` flags.
* **Bug Fixes**
  * Added fail-fast validation to prevent running when **observability** and **security** are both enabled.
* **Testing**
  * Expanded RBAC/JWT integration tests and updated integration log collection to be cluster-wide (per-pod).
* **Chores**
  * Improved local/test setup with configurable Kind config, configurable Helm wait timeouts, safer Keycloak output, and Docker Hub mirror support.

## Kind Integration

Kind integration completed with **success** for [`905b0dd`](https://github.com/NVIDIA/nv-config-manager/commit/905b0dde5ac2946fde4483dba7c950f828a02229) in [this workflow run](https://github.com/NVIDIA/nv-config-manager/actions/runs/30028433853).


src/tests/integration/conftest.py
```diff
@@ -1003,3 +1060,271 @@ def _run(script: str, timeout: int = 20) -> subprocess.CompletedProcess[str]:
...
+    the Nautobot pod, giving tests full ORM access to assert Group /
+    ObjectPermission state without the interactive REPL echoing prompts or the
+    Shell-Plus auto-import banner into stdout. Fails the test on a non-zero exit
+    so ORM errors surface loudly.
+    """
+
+    def _run(script: str, timeout: int = 60) -> str:
+        proc = subprocess.run(
+            [
+                "kubectl",
```
ryanheffernan, line 1371: i'm a bit worried about subprocessing kubectl here - is there any way this could leak and execute against some other local kube context when running locally?

src/tests/integration/test_rbac_group_mapping.py
```diff
@@ -0,0 +1,243 @@
...
+        pytest.skip("group-mapping RBAC tests require --rbac")
+    configmap = f"{rbac_release}-nautobot-group-mapping"
+    res = subprocess.run(
+        ["kubectl", "get", "configmap", "-n", config_manager_namespace, configmap],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if res.returncode != 0:
+        pytest.skip(
```
ryanheffernan: A missing ConfigMap should probably fail rather than skip.

src/tests/integration/test_rbac_group_mapping.py
```diff
@@ -0,0 +1,243 @@
...
+        "    })\n"
+    )
+    return _nbshell_json(nbshell, body)
+
+
+def _managed_perms(nbshell: Callable[[str], str], group_name: str) -> list[str]:
+    """Return the ``<group>_<action>`` ObjectPermission names for *group_name*."""
+    body = (
+        "from nautobot.users.models import ObjectPermission\n"
+        f"names = list(ObjectPermission.objects.filter(name__startswith={group_name + '_'!r})"
```
ryanheffernan, line 119: This only verifies the ObjectPermission names, should also probably verify the `view` and `change` grants so empty or incorrectly scoped permissions cannot pass the integration suite.

src/tests/integration/conftest.py
dmathur0: there's more uses of kubectl already there (port forwarding gathering secrets https://github.com/NVIDIA/nv-config-manager/blob/main/src/tests/integration/conftest.py#L359), we can lock down the contexts a bit more to only use kind? or to add an extra prompt it's not running against kind? or use the kubernetes python library

src/tests/integration/test_rbac_group_mapping.py
dmathur0: i'm pretty sure this should view all grants but let me check

dmathur0: nevermind, it was only names. grants added

src/tests/integration/conftest.py
ryanheffernan: good call, if we already have other examples of this then we can leave this as-is here and then maybe file this as a followup hardening task

dmathur0: Tracked as a follow-up hardening task: #130 (pin fixture `kubectl` calls to a `kind-*` context / guard against ambient-context leakage). Leaving the current code as-is here since there are already pre-existing `kubectl` subprocess usages in this file.

src/tests/integration/test_rbac_group_mapping.py
```diff
@@ -0,0 +1,483 @@
...
+            config_manager_namespace,
+            configmap,
+            "-o",
+            "name",
+            "--ignore-not-found",
+        ],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
```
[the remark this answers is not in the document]
dmathur0: Addressed in e932e36 — the ConfigMap probe now runs with `timeout=30` and fails cleanly on `subprocess.TimeoutExpired` (matching conftest's `_kubectl_run` bound), so the session-scoped fixture can't hang the suite.

daveshm:
https://github.com/NVIDIA/nv-config-manager/actions/runs/29458096628

src/tests/integration/test_rbac_group_mapping.py
```diff
@@ -0,0 +1,504 @@
...
+    assert change is not None, f"missing change perm; got {sorted(grants)}"
+    assert change["actions"] == ["change"], f"change perm has wrong actions: {change['actions']}"
+    # Mapped to ``change: dcim.* + ipam.*`` -> both app scopes must be present.
+    change_ots = change["object_types"]
+    assert any(o.startswith("dcim.") for o in change_ots), (
+        f"change perm missing dcim.* scope: {change_ots}"
+    )
+    assert any(o.startswith("ipam.") for o in change_ots), (
+        f"change perm missing ipam.* scope: {change_ots}"
+    )
```
ryanheffernan, line 274: a grant that has extra scopes after ipam.* would also pass here, should probably check for the exact scopes we want to see

src/tests/integration/test_rbac_group_mapping.py
```diff
@@ -0,0 +1,504 @@
...
+    rbac_api_login: Callable[[str], int],
+    rbac_nbshell: Callable[[str], str],
+) -> None:
+    """An ``is_superuser: true`` mapping entry promotes the user on login."""
+    status = rbac_api_login(USER_ADMIN)
+    assert status == 200, f"expected authorized 200 for {USER_ADMIN}, got {status}"
+
+    state = _user_state(rbac_nbshell, USER_ADMIN)
+    assert state["exists"], f"{USER_ADMIN} was not created by the JWT login"
+    assert state["is_superuser"], f"{USER_ADMIN} was not promoted to superuser: {state}"
```
ryanheffernan: make kind-up-sec already adds a nautobot superuser so this will pass even if the group promotion path is broken - need to reset the feilds to false before the test

src/tests/integration/test_rbac_group_mapping.py
```diff
@@ -0,0 +1,504 @@
...
+        "        'actions': sorted(p.actions),\n"
+        "        'object_types': sorted(\n"
+        "            f'{ct.app_label}.{ct.model}' for ct in p.object_types.all()\n"
+        "        ),\n"
+        "    }\n"
+        "    for p in qs\n"
+        "}\n"
+        "_emit({'grants': out})\n"
+    )
+    return _nbshell_json(nbshell, body)["grants"]
```
ryanheffernan, line 154: this doesn't return the group assignments of ObjectPermissions, so the marker test will pass if the marker exists but is orphaned

.github/workflows/kind-integration.yml
```diff
@@ -174,12 +256,35 @@ jobs:
...
+          TEST_TARGET="${TEST_PATH}"
+
           for attempt in 1 2 3; do
             echo "Attempt ${attempt} of 3"
             results_file="integration-test-results-${attempt}.xml"
             set +e
-            timeout --kill-after=1m 15m uv run pytest "${TEST_PATH}" -v \
+            timeout --kill-after=1m 15m uv run pytest "${TEST_TARGET}" -v \
               --ci \
+              ${EXTRA_FLAGS} \
```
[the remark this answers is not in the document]
dmathur0: Thanks — considered this, but the retry is safe here because the RBAC suite is convergent by design rather than assuming a pristine cluster:

- **Reconcile-on-login is idempotent.** The behavior under test re-reads the ConfigMap mapping and reconciles the user's groups + managed `ObjectPermission`s to match on *every* login (adds missing, overwrites via `update_or_create`, prunes stale `<group>_<action>`/marker records by fingerprint). So a user left half-mutated by a killed attempt is reconciled back to the mapping's intended state on the next attempt's first login — state converges, it doesn't compound.
- **Tests self-seed idempotently.** Preconditions are established with `get_or_create`, and teardown runs inside a `web_request_context` finalizer. Assertions are against the reconciled end-state relative to the deployed mapping, not against an empty DB, so leftovers from a prior attempt are deterministically overwritten/pruned.
- **Retries target infra flakes, not logic.** The loop exists for transient failures (SPIRE/Keycloak port-forward, image pulls, deploy timeouts). Full reset/re-seed per attempt would mean a redeploy on every retry, defeating the cheap-flake purpose, with no correctness gain given the convergence above.

Leaving the loop as-is.

dmathur0:
https://github.com/NVIDIA/nv-config-manager/actions/runs/29763033391

.github/workflows/kind-integration.yml
```diff
@@ -159,12 +162,42 @@ jobs:
...
+          # pulling their images in-cluster through the runner's registry proxy
+          # cache. We deliberately do NOT pre-`kind load` those images: the
+          # runner's Docker uses the containerd image store, whose multi-arch
+          # index `kind load` can't import ("content digest not found"), and the
+          # in-cluster pulls complete within the helm timeout anyway.
+          echo "Default run: kind-up-sec security stack + CONFIGURED group mapping"
+          make kind-up-sec \
             KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME}" \
-            INSTALL_CONFIG=deploy/configs/ci-integration-test.yaml \
             HELM_TIMEOUT=20m
```
ryanheffernan: This 20m timeout doesn't reach `scripts/install-security-dependencies` which runs with `--timeout=5m`

.github/workflows/kind-integration.yml
ryanheffernan: we probably want to expand this to collect logs from all namespaces

.github/workflows/kind-integration.yml
dmathur0: Good catch — fixed in 88c5dd8. Root cause exactly as you said: `HELM_TIMEOUT` only reached the app installer, never `install-security-dependencies`, which hardcoded `--wait --timeout 5m` for cert-manager/Envoy/kgateway/SPIRE. On CI runners SPIRE's in-cluster pulls regularly blew past 5m ("context deadline exceeded").

The script now takes a `--helm-timeout` option (default 5m for local dev) and the Makefile's `kind-up-secure` passes `$(HELM_TIMEOUT)` through, so CI's 20m now actually governs the security-dep installs. Left the fast `spire-crds` install at 2m deliberately (CRDs finish in seconds; a long ceiling there would only mask a genuinely stuck install).

.github/workflows/kind-integration.yml
dmathur0: Done in 88c5dd8 — the Collect step now uses `kubectl get pods -A` / `get events -A` and writes per-pod logs into a per-namespace subdirectory, so a failed security bring-up in spire-system/keycloak/envoy-gateway-system/cert-manager/cnpg-system is captured in the artifact instead of only the app namespace.

.github/workflows/kind-integration.yml
```diff
@@ -159,13 +168,57 @@ jobs:
...
+          # don't touch auth and shouldn't pay that cost.
+          #
+          # `make kind-up-sec` creates the cluster and installs the security
+          # dependencies (cert-manager, Envoy Gateway, SPIRE, Keycloak) itself,
+          # pulling their images in-cluster through the runner's registry proxy
+          # cache. We deliberately do NOT pre-`kind load` those images: the
+          # runner's Docker uses the containerd image store, whose multi-arch
+          # index `kind load` can't import ("content digest not found"), and the
+          # in-cluster pulls complete within the helm timeout anyway.
+          elif [ "${SECURITY}" = "true" ]; then
```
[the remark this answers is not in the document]
dmathur0: Fixed in b1e5678 — added a guard at the top of the Deploy step that fails fast (`exit 2`) when both `observability` and `security` are true, before any deploy branch runs.

.github/workflows/kind-integration.yml
```diff
@@ -181,13 +234,93 @@ jobs:
...
+          #            when no spiffe-helper sidecar pod is present.
+          # The mock topology is loaded by kind-up-sec (deploy/configs/local-sec.yaml
+          # installs development/mock_topology and runs MockTopologyDesign), so the
+          # data-dependent tests have fixtures.
+          #
+          # Default and observability runs use the plain unauthenticated deploy,
+          # so they pass no auth flags.
+          EXTRA_FLAGS=""
+          if [ "${SECURITY}" = "true" ]; then
+            EXTRA_FLAGS="--sso --rbac --spiffe"
```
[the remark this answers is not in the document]
dmathur0: Fixed in b1e5678 — bounded both calls: `--connect-timeout 5 --max-time 10` on the discovery GET and `--connect-timeout 5 --max-time 30` on the token POST, so neither can block indefinitely outside the pytest timeout.

dmathur0:
## Reverted to an opt-in `security` toggle (not always-on)

Heads up on a design change in the latest push: the full security stack (Keycloak + SPIRE) is now **opt-in via a `security: true` workflow input** rather than running on every Kind integration run.

**Why:** the always-on secure stack added **~7 minutes** to every run — the **SPIRE server install is the long pole**. Measured locally on a throwaway kind cluster: `spire-crds` 4s, `spire --wait` **8m57s**; in CI the runner's registry proxy cache compresses the whole security-deps install to roughly **+6–7m net**. Concretely, security-enabled runs land at **~27–28m** vs **~21–22m** for the plain stack. Most PRs don't touch auth, so paying that tax on every run isn't worth it.

**Behavior now** (mirrors the existing `observability` input):
- Default (and `/kind test`) → fast plain `make kind-up`, no auth flags.
- `security: true` → `make kind-up-sec` + CONFIGURED group mapping + the auth-aware suite (`--sso --rbac --spiffe`) with the headless OIDC token mint.
- `observability` and `security` are mutually exclusive (the deploy step now fails fast if both are set).

**Validation:** a 3-run flake sample on the (pre-toggle) always-on config passed **3/3** green — stable once the SPIRE helm-timeout was raised — and a dedicated `security: true` run confirms the opt-in path works end to end.

Makefile
```diff
@@ -867,7 +897,8 @@ kind-up-secure:
 		--keycloak-hostname $(KIND_SEC_KEYCLOAK_HOSTNAME) \
 		--spiffe-trust-domain $(KIND_SEC_SPIFFE_TRUST_DOMAIN) \
 		--keycloak-admin-password $(KIND_SEC_KEYCLOAK_ADMIN_PASSWORD) \
-		--oidc-client-secret $(KIND_SEC_OIDC_CLIENT_SECRET)
+		--oidc-client-secret $(KIND_SEC_OIDC_CLIENT_SECRET) \
```
ryanheffernan, line 881: nit: i know these are just dev credentials but it might be worth ensuring we don't echo raw credentials like this secret or the keycloak password below in case in the future these scripts get run in a CI environment with real creds? not a blocker for this PR

ryanheffernan:
lgtm, assuming it's ready i can approve once the merge commit is in

daveshm:
> lgtm, assuming it's ready i can approve once the merge commit is in

it's very very unstable right now, keeps failing on something new in the new charts, docker image pulls, job timeouts... trying to get it to run 5x no issues

Makefile
dmathur0: Good call — fixed in c0c4e95. `make` echoes recipe lines by default, so the `install-security-dependencies` and `render-local-security-config` invocations were printing the expanded `--keycloak-admin-password` / `--oidc-client-secret` values into the (public) CI log. I `@`-suppressed both commands (kept a short `echo` so the step is still visible in logs). I also caught that the `install-security-dependencies` summary itself echoed the raw admin password on stdout, so that's masked too now. All dev defaults today, but this keeps real secrets out of logs if those vars are ever overridden.

scripts/configure-kind-dockerhub-mirror
```diff
@@ -0,0 +1,118 @@
...
+        printf '%s' "${DOCKERHUB_MIRROR}"
+        return
+    fi
+
+    local buildkit_config="/etc/buildkit/buildkitd.toml"
+    if [ -f "${buildkit_config}" ]; then
+        # Grab the first quoted value on a `mirrors = [ ... ]` line, tolerating
+        # single or double quotes (e.g. mirrors = ['cache.example.com']).
+        sed -nE "s/.*mirrors[[:space:]]*=[[:space:]]*\[?[[:space:]]*[\"']([^\"']+)[\"'].*/\1/p" \
+            "${buildkit_config}" 2>/dev/null | head -n1
```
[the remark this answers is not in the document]
dmathur0: Valid — fixed in e325435. `discover_mirror` now parses the mirror only from the `[registry."docker.io"]` section (awk gates on that header and resets on the next `[`), instead of grabbing the first `mirrors =` line anywhere. Still tolerates single- and double-quoted values.

The downstream risk you flagged is exactly why this matters: a non-empty result disables the Makefile's preload fallback, so misassigning another registry's endpoint to docker.io would have broken secure-stack pulls with no backstop. Verified against a multi-section `buildkitd.toml` (ghcr.io first, then docker.io, then quay.io) — it correctly returns the docker.io endpoint, and a config with no docker.io section returns empty so we fall back to the preload.

scripts/configure-kind-dockerhub-mirror
```diff
@@ -0,0 +1,130 @@
...
+    if [ -f "${buildkit_config}" ]; then
+        # Extract the mirror endpoint ONLY from the [registry."docker.io"]
+        # section. buildkitd.toml can define mirrors for several registries, so
+        # grabbing the first `mirrors =` anywhere would misassign another
+        # registry's endpoint to docker.io -- and since the secure run relies on
+        # this mirror as its sole backstop, a wrong or unreachable endpoint would
+        # break every in-cluster docker.io pull. Tolerates single- or
+        # double-quoted values.
+        awk '
+            /^\[registry\."docker\.io"\][[:space:]]*$/ { in_docker_hub = 1; next }
```
ryanheffernan, line 77: nonblocking nit: we should accept both `[registry."docker.io"]` and `[registry.'docker.io']` just in case this changes in the future and someone uses a different quote style

ryanheffernan:
i think once we get a good run of observability test to confirm those still work with security disabled this is g2g

dmathur0:
https://github.com/NVIDIA/nv-config-manager/actions/runs/30028433853 run with observability on only

ryanheffernan (approved):

PR #128, opened 2026-07-14
Update helm chart to support image digests

jojung1:
## Description

Adds optional content-addressed image pinning to the Helm chart. Introduces a `nv-config-manager.image` template helper that renders `repository@sha256:...` when a `digest` is set on an image, and otherwise falls back to the existing `repository:tag` behavior (with the current `imageTag` chart-version fallback unchanged).

- New helper `nv-config-manager.image` in `templates/_helpers.tpl`; digest takes precedence over tag, and an empty `digest: ""` is falsy so it is a no-op.
- Converts the `image:` references for the **6 built service images** (`nvConfigManager`, `nvConfigManagerUi`, `kea`, `keaAdmin`, `nautobot`, `natsReady`) to use the helper. Third-party images (busybox, redis, nats, temporal*, kubectl, etc.), oauth2-proxy, the spiffe/nginx sidecars, and render template plugins are intentionally left as-is — they are version-pinned by hand, not by the release pipeline.
- Adds a commented `digest: ""` default to those 6 images in `values.yaml`.
- `ui.yaml`'s `APP_VERSION` env var intentionally stays on `imageTag` so the UI's `/api/config` and `/api/health` continue to report a human-readable version rather than a digest.
- Adds an `Image digest pin render` step to the existing `helm` job in `public-ci.yml` that exercises both paths (unset digest → tag; set digest → `@sha256:...`).

This is groundwork for a GitOps deploy flow that pins images by digest; it has no effect until a caller sets a digest (e.g. via `--set global.images.<name>.digest=...` or a values override), so it is safe to merge independently.

## Validation

Covered by CI: the `Image digest pin render` step in the `helm` job asserts (a) a default render pins no image by digest — the tag fallback still applies — and (b) setting a digest on each of the 6 built services renders it as `repository@sha256:...`. Reproduce locally:

```bash
# 1) Byte-identical render when no digest is set (adoption safety).
#    Rendered against real test values and diffed against the
#    pre-change chart — no differences.
helm template nv-config-manager deploy/helm -f <test values> > after.yaml
diff before.yaml after.yaml        # empty

# 2) Digest wins when set.
helm template nv-config-manager deploy/helm -f <test values> \
  --set global.images.nvConfigManager.digest=sha256:aaaa... \
  --set global.images.kea.digest=sha256:bbbb... \
  | grep 'image:'
#   -> registry.example.com/nvidia/nv-config-manager@sha256:aaaa...
#   -> registry.example.com/nvidia/nv-config-manager-kea@sha256:bbbb...

# 3) Lint clean.
helm lint deploy/helm -f <test values>
```
- [x] Standard CI passes.

The kind integration test is manual due to taking ~30 min to complete. When the PR is ready for
review, approve the current commit and start the suite with these PR comments:

```text
/ok to test <sha>
/kind test
```

only while debugging.

Kind integration completed with **success** for [`11b460eba1c5`](https://github.com/NVIDIA/nv-config-manager/actions/runs/29362683540).

## Checklist

## Summary

- **New Features**
  - Added support for pinning container images by digest, improving deployment consistency.
  - Updated Config Store, DHCP, MCP, Nautobot, Render Service, Temporal, UI, and ZTP components to resolve images via a shared, consistent image reference helper.
  - When a digest is configured, it takes precedence over tag-based image selection; tag fallback remains available.
- **Tests / CI**
  - Enhanced public CI with a render-time check to verify digest pinning behavior and ensure tag references are not used when digests are set.

### DIFF
--- a/.github/workflows/public-ci.yml
+++ b/.github/workflows/public-ci.yml
@@ -289,6 +289,47 @@ jobs:
               if (sc == 0 || sc != ok) exit 1
             }
           ' /tmp/nvcm-probes.yaml
+      - name: Image digest pin render
+        run: |
+          # global.images.<name>.digest must pin built-service images by digest
+          # (repository@sha256:...) and win over the tag; an unset digest must
+          # fall back to repository:tag. Render-time check only (no cluster) so
+          # it stays out of the slow kind-integration path.
+          set -euo pipefail
+          names="nvConfigManager nvConfigManagerUi kea keaAdmin nautobot natsReady"
+          d_nvConfigManager=sha256:1111111111111111111111111111111111111111111111111111111111111111
+          d_nvConfigManagerUi=sha256:2222222222222222222222222222222222222222222222222222222222222222
+          d_kea=sha256:3333333333333333333333333333333333333333333333333333333333333333
+          d_keaAdmin=sha256:4444444444444444444444444444444444444444444444444444444444444444
+          d_nautobot=sha256:5555555555555555555555555555555555555555555555555555555555555555
+          d_natsReady=sha256:6666666666666666666666666666666666666666666666666666666666666666
+
+          # Default render (no digests set) must not pin any image by digest.
+          helm template test deploy/helm/ --values deploy/helm/values-ci.yaml >/tmp/nvcm-nodigest.yaml
+          if grep -Eq 'image:[[:space:]]*"?[^"[:space:]]+@sha256:' /tmp/nvcm-nodigest.yaml; then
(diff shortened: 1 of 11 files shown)


ryanheffernan:
worth adding a quick helm ci test to the existing suite to exercise the digest path. 

otherwise lgtm!

jojung1:
@ryanheffernan added helm digest testing to public-ci.yaml. If it looks good we can go ahead and get this merged 👍

ryanheffernan (approved):

PR #133, opened 2026-07-15
feat(spx): support removing port assignments and cleaning up VRFs

ryanheffernan:
## Description

Allow the SpX Overlay Assignment and Tenant Change workflows to remove selected port assignments without requiring a replacement overlay.

Previously, an overlay was required, so the workflow could only move or create assignments. It also retained the device/VRF association after the last interface stopped mapping to that VRF. This change:

- treats an omitted overlay ID as a removal-only operation
- allows the tenant-change UI to submit selected ports with no replacement overlay
- clears the selected interfaces' VRF and Spectrum-X overlay assignments
- removes stale device-level overlay and device/VRF associations when no device interface still uses them
- in order to support the partial deploy workflow for removal-only deploys, add a `use_full_intended_config` parameter to the tenant deployment workflow. Without this, the diff doesn't report anything for 'unset only' diffs and we don't do a deploy. This parameter ensures a full config diff occurs to determine if a deploy should happen, but we still restrict the deploy to the strict set of allowed commands for the spx workflows.

This PR also makes the tenant deploy workflow internal only - it should only ever be called by the Spx Tenant Change workflow.

## Validation

- [x] Standard CI passes.
- [x] Kind integration passes, or this PR explains why it was not run. Not run locally; approve the current commit and run the manual suite before merge.

The kind integration test is manual due to taking ~30 min to complete. When the PR is ready for
review, approve the current commit and start the suite with these PR comments:

```text
/ok to test <sha>
/kind test
```

only while debugging.

Kind integration completed with **success** for [`0f7a885`](https://github.com/NVIDIA/nv-config-manager/commit/0f7a885463a7029737f7f62568abc0e82d74300e) in [this workflow run](https://github.com/NVIDIA/nv-config-manager/actions/runs/31726968880).

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Spectrum-X overlay assignments can now be removed by leaving the Overlay ID blank.
  * Tenant changes support assignment or removal, with clearer results, cleanup, synchronization, and deployment reporting.
  * Port selections require at least one valid port.
  * Deployments can apply the full intended configuration when needed.
* **Documentation**
  * Updated overlay assignment and tenant-change guides with removal workflows and verification steps.
* **Breaking Changes**
  * Removed the public Tenant Deploy workflow, endpoint, user interface, and documentation.


docs/user-guides/vpc-tenant-change/index.mdx
```diff
@@ -4,7 +4,7 @@ sidebar-title: SpX Overlay Tenant Change
...
+The SpX Overlay Tenant Change workflow drives an end-to-end overlay change against a single device: assign or remove the overlay on selected ports, re-render the tenant configuration, wait for the new render to land in the Config Store, and deploy the rendered configuration to the device. It is the workflow most operators interact with when adding, moving, removing, or expanding a tenant's footprint on a switch — the underlying [SpX Overlay Assignment](../vpc-assignment/index.mdx), [Render Service](../../render/index.mdx), and [Tenant Deploy](../tenant-deploy/index.mdx) workflows are exposed individually for advanced cases but are normally orchestrated by this workflow.
 
 <Warning title="Experimental">
```
zsblevins: we can drop experimental here since this was a pre-QA flag

src/nv_config_manager/temporal/api/workflow_v1.py
```diff
@@ -791,7 +792,11 @@ async def get_workflows(  # pylint: disable=R0913,R0914
 async def get_workflow_types() -> list[str]:
     """Return registered workflow type names."""
     return sorted(
-        [wf.__name__ for wf in NGC_REGISTERED_WORKFLOWS + HELLO_WORLD_REGISTERED_WORKFLOWS]
+        [
```
zsblevins: If this results in changes to the workflow types endpoint, I _think_ it will affect the ability to filter on the child workflows on the workflow table. We can still expose the metadata for them, we just don't want to accept direct invocations of them.

zsblevins (approved):

src/nv_config_manager/temporal/api/workflow_v1.py
ryanheffernan: accidentally included, removed

PR #134, opened 2026-07-15
Add secret-free build stage for untrusted PRs

jojung1:
## Description

Adds the **untrusted, secret-free build stage** for the test/test01 GitOps promote flow. It builds a PR's artifacts on the mirror's `pull-request/<n>` ref (the vetted copy-pr-bot snapshot) with **no protected variables in scope**, and hands them off as GitLab job artifacts for a later protected promote pipeline to publish. This is the first half of a build/promote split so untrusted PR code never runs with secrets.

What this adds:
- **`.gitlab/ci/pr-build.yml`** — two jobs, gated to `pull-request/*` refs on deliberately-triggered sources only (`pipeline|trigger|api|web`, never `push`):
  - `pr-build-image` — a 6-way matrix that builds each service image with no `docker login`, writes it to a local tarball (`docker buildx --output type=docker,dest=…`), gzips it, and uploads it as an artifact.
  - `pr-build-chart` — runs `helm dependency build` + `helm package` (the steps that process untrusted chart input, e.g. fetching subcharts from PR-controlled `Chart.yaml` URLs) and uploads the `.tgz`. Doing this here keeps that untrusted processing out of the credentialed promote pipeline.
- **`Makefile`** — new `DOCKER_BUILD_OUTPUT ?= --push`, used by the six `docker-build-single-*` targets. Defaults to `--push`, so every existing caller (release/main builds) is unchanged; the secret-free build overrides it to emit a tarball instead.
- **`.gitlab/ci/common.yml`** — a workflow rule admitting `pull-request/*` pipelines only from non-push sources. The existing "no push pipelines on non-default branches" guard is untouched.
- **`.gitlab/ci/scripts/pr_ref_guard.sh`** — stale-HEAD guard: verifies the SHA being built is still HEAD of the mirror's `pull-request/<n>` and cross-checks GitHub's copy of that vetted branch (catches mirror lag). Runs before building and again before uploading artifacts.

Security model: `pull-request/*` refs are unprotected, so GitLab withholds all protected variables from these jobs — the build genuinely cannot see registry/push tokens. Nothing here auto-runs; the build is only started deliberately (by the promote pipeline in a follow-up PR, or a manual run). Safe to merge on its own: it changes no existing pipeline behavior and adds no job that runs on push/MR/tag pipelines.

Follow-up PR adds the protected promote pipeline that triggers this build, publishes the artifacts, and commits deploy-state to the ArgoCD repo.

## Validation

- `yq`/yamllint parse of `pr-build.yml`, `common.yml`, and `.gitlab-ci.yml`; `bash -n` on `pr_ref_guard.sh`.
- Confirmed the workflow rule's source allowlist matches the pr-build job rules (`pipeline|trigger|api|web`; `push` excluded on both).
- `Makefile`: verified the change is a no-op for existing callers — `DOCKER_BUILD_OUTPUT` defaults to `--push`, so release/main image builds are byte-for-byte the same invocation.
- `pr-build-chart` logic exercised locally: `helm dependency build` + `helm package` with a `0.0.0-pr<n>.<sha>` dev version produces a valid chart `.tgz`.
- End-to-end pipeline run is exercised by manually running a pipeline on a `pull-request/<n>` ref on the mirror (source `web`) — requires the mirror to replicate `pull-request/*` and the artifact-size limit to be sufficient. Left for the integrated validation once the promote PR lands, unless run manually here.

- [x] Standard CI passes.

Kind integration not run — not applicable. This PR only adds GitLab CI job definitions, a backwards-compatible Makefile flag, and a shell guard; it changes no application code, no Helm chart runtime output, and nothing the Kind integration suite exercises (which runs the app in-cluster via GitHub Actions, not the GitLab pipeline this PR touches).

## Checklist

- [x] Generated artifacts are updated when applicable, such as OpenAPI specs, docs screenshots, or Helm/rendered outputs.

## Summary by CodeRabbit

* **New Features**
  * Added an untrusted pull-request CI stage that builds secret-free container image and Helm chart artifacts for later promotion.
* **Bug Fixes**
  * Improved pull-request pipeline triggering so pipelines run only for vetted pipeline sources.
  * Added safeguards to prevent artifact creation/upload for stale pull-request refs.
* **Chores**
  * Made single-arch Buildx output configurable (defaults to pushing; can export tar artifacts) and updated messaging.
  * Refactored single-arch Temporal image builds into component targets to control output behavior.

### DIFF
--- a/.gitlab/ci/pr-build.yml
+++ b/.gitlab/ci/pr-build.yml
@@ -45,9 +45,6 @@ pr-build-image:
     DOCKER_HOST: tcp://docker:2375/
     DOCKER_TLS_CERTDIR: ""
     DOCKER_DRIVER: overlay2
-    # Untrusted stage: no outbound GitHub calls (rate-limit-prone, and the
-    # authoritative freshness/provenance checks live in the promote consumer).
-    NVCM_PR_GUARD_SKIP_GITHUB: "1"
   rules: *pr-build-rules
   parallel:
     matrix:
@@ -117,9 +114,6 @@ pr-build-chart:
     - docker
   needs: []
   image: $NVCM_CI_IMAGE_ALPINE
-  variables:
-    # Untrusted stage: no outbound GitHub calls (see pr-build-image).
-    NVCM_PR_GUARD_SKIP_GITHUB: "1"
   rules: *pr-build-rules
   before_script:
     - apk add -q bash git curl helm yq
--- a/.gitlab/ci/pr-build.yml
+++ b/.gitlab/ci/pr-build.yml
@@ -62,6 +62,12 @@ pr-build-image:
         NVCM_PR_IMAGE: nv-config-manager-nautobot
       - NVCM_PR_TARGET: nats-ready
         NVCM_PR_IMAGE: nv-config-manager-nats-ready
+      - NVCM_PR_TARGET: temporal-server
+        NVCM_PR_IMAGE: nv-config-manager-temporal
+      - NVCM_PR_TARGET: temporal-bootstrap
+        NVCM_PR_IMAGE: nv-config-manager-temporal-bootstrap
+      - NVCM_PR_TARGET: temporal-ui
+        NVCM_PR_IMAGE: nv-config-manager-temporal-ui
   before_script:
     - apk add -q make bash git curl coreutils
     - docker info
--- a/.gitlab/ci/pr-build.yml
+++ b/.gitlab/ci/pr-build.yml
@@ -31,8 +31,6 @@
 # uploads, and lets the promote side download by job id.
 pr-build-image:
   stage: build
-  # Restricted, isolated runners for untrusted PR builds (no privileged access
-  # to the internal DCA fleet). Protected promote jobs stay on the dca runners.
   tags:
     - dgxc-corp-restricted
     - docker
--- a/.gitlab/ci/pr-build.yml
+++ b/.gitlab/ci/pr-build.yml
@@ -31,10 +31,12 @@
 # uploads, and lets the promote side download by job id.
 pr-build-image:
   stage: build
+  # Restricted, isolated runners for untrusted PR builds (no privileged access
+  # to the internal DCA fleet). Protected promote jobs stay on the dca runners.
   tags:
-    - dca
-    - linux/amd64
+    - dgxc-corp-restricted
     - docker
+    - linux/amd64
   needs: []
   image: $NVCM_CI_IMAGE_DOCKER
   services:
@@ -108,10 +110,12 @@ pr-build-image:
 # from PR source.
 pr-build-chart:
   stage: build
+  # Restricted, isolated runners for untrusted PR builds (no privileged access
+  # to the internal DCA fleet). Protected promote jobs stay on the dca runners.
   tags:
-    - dca
-    - linux/amd64
+    - dgxc-corp-restricted
     - docker
+    - linux/amd64
   needs: []
   image: $NVCM_CI_IMAGE_ALPINE
   rules: *pr-build-rules
--- a/.gitlab/ci/pr-build.yml
+++ b/.gitlab/ci/pr-build.yml
@@ -45,6 +45,9 @@ pr-build-image:
     DOCKER_HOST: tcp://docker:2375/
     DOCKER_TLS_CERTDIR: ""
     DOCKER_DRIVER: overlay2
+    # Untrusted stage: no outbound GitHub calls (rate-limit-prone, and the
+    # authoritative freshness/provenance checks live in the promote consumer).
+    NVCM_PR_GUARD_SKIP_GITHUB: "1"
   rules: *pr-build-rules
   parallel:
     matrix:
@@ -114,6 +117,9 @@ pr-build-chart:
     - docker
   needs: []
   image: $NVCM_CI_IMAGE_ALPINE
+  variables:
+    # Untrusted stage: no outbound GitHub calls (see pr-build-image).
+    NVCM_PR_GUARD_SKIP_GITHUB: "1"
   rules: *pr-build-rules
   before_script:
     - apk add -q bash git curl helm yq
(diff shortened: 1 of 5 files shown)


.gitlab/ci/pr-build.yml
```diff
@@ -0,0 +1,144 @@
...
+  - when: never
+
+# One job per image: bounds each artifact's size, parallelizes the builds and
+# uploads, and lets the promote side download by job id.
+pr-build-image:
+  stage: build
+  tags:
+    - dca
+    - linux/amd64
+    - docker
```
ryanheffernan, line 36: for an unpriveleged build we probably don't want to use the internal priveleged runners. the gitlab [docs](https://docs.gitlab.com/runner/security/) recommend isolated, ephemeral machines for untrusted workloads so we probably need to look in to something like that

.gitlab/ci/scripts/pr_ref_guard.sh
```diff
@@ -0,0 +1,67 @@
...
+# on a pull-request ref calls this to confirm the ref still exists on the
+# mirror and still points at the SHA being built/promoted, with a best-effort
+# cross-check against upstream GitHub.
+#
+# Usage: pr_ref_guard.sh [ref] [expected-sha]
+#   ref          defaults to $CI_COMMIT_REF_NAME
+#   expected-sha defaults to $CI_COMMIT_SHA
+#
+# Uses only $CI_REPOSITORY_URL (embeds the read-only job token, available on
+# unprotected refs) and unauthenticated GitHub API access - no secrets.
```
ryanheffernan: `CI_JOB_TOKEN` isn't just read-only, can we do this in an environment without any access to internal registries or tokens?

.gitlab/ci/pr-build.yml
```diff
@@ -0,0 +1,144 @@
...
+        NVCM_PR_IMAGE: nv-config-manager-ui
+      - NVCM_PR_TARGET: nb
+        NVCM_PR_IMAGE: nv-config-manager-nautobot
+      - NVCM_PR_TARGET: nats-ready
+        NVCM_PR_IMAGE: nv-config-manager-nats-ready
+  before_script:
+    - apk add -q make bash git curl coreutils
+    - docker info
+    # Fail fast if the PR moved or closed while this pipeline was queued.
+    - bash .gitlab/ci/scripts/pr_ref_guard.sh
```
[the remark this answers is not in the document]
ryanheffernan: this sounds pretty legit - we should probably find a way around this. similar to my suggestion above about the token env variables

.gitlab/ci/scripts/pr_ref_guard.sh
```diff
@@ -0,0 +1,67 @@
...
+        fi
+        ;;
+    404)
+        echo "ERROR: ${ref} does not exist on GitHub (PR closed/merged or copy-pr-bot"
+        echo "branch removed); the mirror copy is an orphaned stale branch. Refusing."
+        exit 1
+        ;;
+    *)
+        echo "WARN: could not verify ${ref} against GitHub (HTTP ${github_status:-000}: network, rate limit, or API error); mirror check passed."
+        ;;
```
ryanheffernan: this pipeline makes 14 unauthenticated GitHub API calls, while GitHub limits unauthenticated clients to 60 requests/hour per source IP. If our runners share a NAT address, github may rate-limit and respond with 403 or 429 which will cause this script to warn and succeed. 

I'd suggest to move the authoritative comparison to the protected promotion consumer where we can make authenticated calls, and severely limit or eliminate unauthenticated calls.

.gitlab/ci/pr-build.yml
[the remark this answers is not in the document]
ryanheffernan: the pull-request branch is controlled by copy-pr-bot and only created when a trusted maintainer triggers it, so it isn't 'fully attacker controlled'

.gitlab/ci/scripts/pr_ref_guard.sh
jojung1: `CI_JOB_TOKEN` will always be accessible because this get injected into every gitlab job regardless of the environment. We can restrict the permissions of it via gitlab project settings, which it seems like it's already at its least permissible state (based on what I am seeing).

For internal registries, I am not sure what exact permissions DCA runners provide, but to be sure, we could (and probably should) set up our own ephemeral + isolated runners

.gitlab/ci/scripts/pr_ref_guard.sh
```diff
@@ -0,0 +1,80 @@
...
+# GitHub cross-check: best-effort detection of the mirror lagging behind
+# copy-pr-bot's pull-request/<n> branch. It does NOT compare against the PR's
+# live HEAD (that divergence is surfaced in promote_build_pipeline.sh).
+#
+# Skipped in the untrusted build stage (NVCM_PR_GUARD_SKIP_GITHUB=1): those jobs
+# run many times per pipeline, and unauthenticated GitHub calls behind a shared
+# NAT quickly exhaust the 60 req/hour limit (then degrade to warn-and-succeed).
+# The authoritative freshness + provenance checks run in the protected promote
+# consumer, which can make authenticated calls.
+if [ "${NVCM_PR_GUARD_SKIP_GITHUB:-0}" = "1" ] || [ "${NVCM_PR_GUARD_SKIP_GITHUB:-}" = "true" ]; then
```
ryanheffernan: is there a reason to keep the github calls at all? i.e. can we just delete the code, or is this script meant to be shared with both the unprotected and protected jobs?

jojung1: It was meant to be a fail-fast check originally but that's probably not a strong enough reason to keep the calls. The promote consumer makes one check and shoots a warning on divergence anyway. I'll remove the gh calls.

.gitlab/ci/pr-build.yml
jojung1: Updated the runners to use DCA's new restricted gitlab runners.

ryanheffernan (approved):

PR #143, opened 2026-07-17
Add protected test/test01 promote pipeline

jojung1:
## Description

> **Draft — stacked on #134.** This is the protected promote half of the test/test01 GitOps deploy flow; #134 (the untrusted build stage) must merge first. Kept as a draft until then. The kiwi-argocd ApplicationSet/values changes are a separate MR.

Adds the **protected promote pipeline** that deploys a PR to the shared **test / test01** environments by consuming the immutable artifacts built in #134 and committing a machine-written `deploy-state.yaml` to the env branch in the ArgoCD values repo. It targets **only** test/test01 (every rule requires `NVCM_PROMOTE_ENV` ∈ {test, test01}); production stays on the tag-driven release flow.

Run on the mirror's default branch with `NVCM_PROMOTE_PR` + `NVCM_PROMOTE_ENV`. Jobs (`.gitlab/ci/promote-test-envs.yml`):
- `test-promote-build` — resolves the vetted `pull-request/<n>` snapshot, reuses or triggers the #134 build, and gates provenance on **GitLab pipeline metadata** (sha/ref/status/source) — never on anything the untrusted build wrote.
- `test-promote-push-images` — downloads the image tarballs by job id, `docker load`, retag to `0.0.0-pr<n>.<sha>`, push, capture registry digests.
- `test-promote-chart` — downloads the chart `.tgz` packaged in #134's secret-free stage, validates the render against the env's baseline+overrides+digests, and publishes it (no PR checkout, no repackaging — `helm template` only, which is sandboxed).
- `test-promote-deploy-state` — commits `deploy-state.yaml` to the env branch (`NV_CONFIG_MANAGER_VALUES_PUSH_TOKEN`), owning the GitLab Environment.
- `test-rollback-env` / `test-release-env` — commit a prior deploy-state / reset to canonicals.

**Security model:** the promote stage operates only on already-built immutable artifacts, so no untrusted PR code executes with secrets. Provenance is established from GitLab metadata; the build stage (#134) holds no secrets.

**Default-branch job guards:** a `web`-source pipeline on the default branch with `NVCM_PROMOTE_ENV` set would otherwise also match `docker-build-main-*`, `test-main-*`, sonar, and pulse (their `rules:changes` evaluate true on non-push pipelines). Each gets a leading `- if: $NVCM_PROMOTE_ENV / when: never` so a promote run doesn't rebuild/retest main as a side effect. Normal push/MR/tag pipelines are unaffected.

**Chart source:** the promote publishes the dev chart to the existing **NGC Helm registry** (`https://helm.ngc.nvidia.com/nvidian/cfa`) via the `ngc` target; ArgoCD pulls it by exact version. (Confirmed exact-prerelease pins resolve there.)

**Required protected CI/CD variables:** `NVCM_MIRROR_API_TOKEN`, `NVCM_TEST_ENV_TARGETS`, `NVCM_CHART_REPO`, `NV_CONFIG_MANAGER_VALUES_PUSH_TOKEN`, `NVCM_VALUES_REPO_PATH`, plus the existing image/chart target vars. Documented in `.gitlab/ci/README.md`.

## Validation

- `yq` parse of `.gitlab-ci.yml` + `promote-test-envs.yml` + the four guarded CI files; `bash -n` on all four promote scripts.
- Confirmed the promote source allowlist and default-branch guards behave as intended.
- Exercised the build→promote chart handoff locally: packaged the chart as #134's `pr-build-chart` does, then ran the promote render gate against the real test baseline+overrides+digests → renders `nvcr.io/nvidian/cfa/...@sha256:...`.
- End-to-end pipeline validation is gated on #134 merging (nothing here runs until the build stage exists) and the protected variables being provisioned.

- [x] Standard CI passes.

Kind integration not run — not applicable. This PR only adds GitLab CI job definitions and shell scripts for the deploy pipeline; it changes no application code or Helm chart runtime output, and nothing the Kind suite exercises.

## Checklist

- [ ] Generated artifacts are updated when applicable, such as OpenAPI specs, docs screenshots, or Helm/rendered outputs.

## Summary by CodeRabbit

* **New Features**
  * Strengthened **test/test01** promotion to use file-artifact-based provenance, validate chart integrity (including digest-verified rendering and mismatch detection), and pin deployments to attested chart versions and image digests.
  * Added a fail-closed test-environment configuration resolver restricted to `test` and `test01`.
* **Documentation**
  * Updated the promotion runbook with downstream values-repo token allowlist requirements and token usage rules.
* **Bug Fixes**
  * Prevented promote pipelines from triggering default-branch image rebuilds, manifest publishing, Sonar scans, Pulse scans, and related test jobs; tightened build/pipeline execution guards.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,3 +10,4 @@ include:
   - local: .gitlab/ci/release.yml
   - local: .gitlab/ci/airgapped.yml
   - local: .gitlab/ci/pr-build.yml
+  - local: .gitlab/ci/promote-test-envs.yml
--- a/.gitlab/ci/promote-test-envs.yml
+++ b/.gitlab/ci/promote-test-envs.yml
@@ -62,6 +62,13 @@ test-promote-build:
   artifacts:
     reports:
       dotenv: promote.env
+    # Also a plain file: downstream jobs read the provenance-critical values
+    # (resolved PR ref/SHA, build job ids) from disk, since pipeline variables
+    # outrank dotenv and could otherwise redirect artifact downloads or defeat
+    # the stale-HEAD guard.
+    paths:
+      - promote.env
+    expire_in: 1 week
 
 # -----------------------------------------------------------------------------
 # 2) Download image tars by job id, load, retag, push, capture digests.
@@ -158,7 +165,10 @@ test-promote-chart:
       # promote_build_pipeline.sh). This job never checks out or packages PR
       # source, so no untrusted chart code (e.g. `helm dependency build`) runs
       # in this credentialed context.
-      : "${CHART_BUILD_JOB_ID:?missing dotenv from test-promote-build}"
+      # Read from the file artifact, not dotenv: an overriding pipeline variable
+      # could otherwise point this download at a different job's artifacts.
+      CHART_BUILD_JOB_ID="$(grep -m1 '^CHART_BUILD_JOB_ID=' "${CI_PROJECT_DIR}/promote.env" | cut -d= -f2-)"
+      [ -n "$CHART_BUILD_JOB_ID" ] || { echo "❌ CHART_BUILD_JOB_ID missing from promote.env"; exit 1; }
       # SECURITY: this archive is produced by the untrusted build stage - the PR
       # controls pr-build.yml, and therefore the artifact's paths and contents.
       # Extract into a throwaway directory, NEVER into CI_PROJECT_DIR: unzipping
@@ -173,7 +183,12 @@ test-promote-chart:
 
       # Provenance: require exactly the chart version this promote run computed
       # from trusted PR metadata. A chart stamped with any other version does
-      # not match and is rejected (fails closed).
+      # not match and is rejected (fails closed). Read the version from the
+      # build job's FILE artifact for the same reason as the job id above -
+      # push-images already tagged the images from that same source, so taking
+      # it from (overridable) dotenv here could desynchronise the two.
+      PROMOTE_VERSION="$(grep -m1 '^PROMOTE_VERSION=' "${CI_PROJECT_DIR}/promote.env" | cut -d= -f2-)"
+      [ -n "$PROMOTE_VERSION" ] || { echo "❌ PROMOTE_VERSION missing from promote.env"; exit 1; }
       CHART_NAME=nv-config-manager
       chart_file="${chart_dl}/x/chart/${CHART_NAME}-${PROMOTE_VERSION}.tgz"
       if [ ! -f "$chart_file" ]; then
@@ -208,7 +223,7 @@ test-promote-chart:
       fi
       git clone "$values_repo_url" /tmp/values-repo
       # Capture the exact main SHA whose baseline values this render validates.
-      # write_deploy_state consumes it (via dotenv) instead of re-resolving
+      # write_deploy_state reads it from this file instead of re-resolving
       # origin/main, so the DEPLOYED baseline is the one that was validated even
       # if main moves before the deploy-state job runs.
       baseline_revision="$(git -C /tmp/values-repo rev-parse HEAD)"
@@ -229,6 +244,28 @@ test-promote-chart:
       echo "PROMOTE_VERSION=${PROMOTE_VERSION}" >> "${CI_PROJECT_DIR}/chart.env"
       overrides_file="/tmp/values-env/${NVCM_ENV_STATE_DIR}/values-aws-${NVCM_ENV}.overrides.yaml"
       test -f "$overrides_file" || (echo "❌ Missing overrides file ${overrides_file} on ${NVCM_ENV_BRANCH}" && exit 1)
+
+      # Digest pins come from the push job's FILE artifact, not its dotenv
+      # export. write_deploy_state.sh pins the file's values, so reading the
+      # same source here keeps "what we validated" and "what we deploy"
+      # identical even if a same-named pipeline variable is set.
+      digest_attest="${CI_PROJECT_DIR}/digests.env"
+      [ -f "$digest_attest" ] || { echo "❌ Missing ${digest_attest} (artifact from test-promote-push-images)"; exit 1; }
+      attest_digest() {
+        local key="$1" val
+        val="$(grep -m1 "^${key}=" "$digest_attest" | cut -d= -f2- || true)"
+        [ -n "$val" ] || { echo "❌ ${key} missing from digests.env"; exit 1; }
+        printf '%s' "$val"
+      }
+      DIGEST_NV_CONFIG_MANAGER="$(attest_digest DIGEST_NV_CONFIG_MANAGER)"
+      DIGEST_NV_CONFIG_MANAGER_UI="$(attest_digest DIGEST_NV_CONFIG_MANAGER_UI)"
+      DIGEST_NV_CONFIG_MANAGER_KEA="$(attest_digest DIGEST_NV_CONFIG_MANAGER_KEA)"
+      DIGEST_NV_CONFIG_MANAGER_KEA_ADMIN="$(attest_digest DIGEST_NV_CONFIG_MANAGER_KEA_ADMIN)"
+      DIGEST_NV_CONFIG_MANAGER_NAUTOBOT="$(attest_digest DIGEST_NV_CONFIG_MANAGER_NAUTOBOT)"
+      DIGEST_NV_CONFIG_MANAGER_NATS_READY="$(attest_digest DIGEST_NV_CONFIG_MANAGER_NATS_READY)"
+      DIGEST_NV_CONFIG_MANAGER_TEMPORAL="$(attest_digest DIGEST_NV_CONFIG_MANAGER_TEMPORAL)"
+      DIGEST_NV_CONFIG_MANAGER_TEMPORAL_BOOTSTRAP="$(attest_digest DIGEST_NV_CONFIG_MANAGER_TEMPORAL_BOOTSTRAP)"
+      DIGEST_NV_CONFIG_MANAGER_TEMPORAL_UI="$(attest_digest DIGEST_NV_CONFIG_MANAGER_TEMPORAL_UI)"
 
       # Render with identical inputs for any package, so the locally validated
       # one and the published one can be compared.
@@ -327,7 +364,14 @@ test-promote-deploy-state:
     - git config --global user.name "NVIDIA Config Manager CI"
   script:
     # Final stale-HEAD check before mutating the env branch.
-    - bash .gitlab/ci/scripts/pr_ref_guard.sh "$PR_REF" "$PR_SHA"
+    # Guard inputs come from the file artifact, not dotenv - an overriding
+    # pipeline variable could otherwise make this check compare against
+    # operator-chosen values and pass trivially.
+    - |
+      pr_ref="$(grep -m1 '^PR_REF=' "${CI_PROJECT_DIR}/promote.env" | cut -d= -f2-)"
+      pr_sha="$(grep -m1 '^PR_SHA=' "${CI_PROJECT_DIR}/promote.env" | cut -d= -f2-)"
+      [ -n "$pr_ref" ] && [ -n "$pr_sha" ] || { echo "❌ PR_REF/PR_SHA missing from promote.env"; exit 1; }
+      bash .gitlab/ci/scripts/pr_ref_guard.sh "$pr_ref" "$pr_sha"
     - |
       env_config="$(bash .gitlab/ci/scripts/test_env_config.sh "$NVCM_PROMOTE_ENV")" || exit $?
       eval "$env_config"
--- a/.gitlab/ci/promote-test-envs.yml
+++ b/.gitlab/ci/promote-test-envs.yml
@@ -178,14 +178,20 @@ test-promote-chart:
       # Render gate against the env's real values layering (pinned baseline +
       # env-branch overrides + digest pins) BEFORE publishing or touching the
       # env branch. helm template is sandboxed (no env access), so rendering the
-      # untrusted chart with the values token present cannot leak secrets; the
-      # packaged .tgz already vendors its subcharts, so nothing is fetched.
+      # untrusted chart with this job's credentials present cannot leak them;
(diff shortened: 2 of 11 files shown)


.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,365 @@
...
+    DOCKER_DRIVER: overlay2
+  rules: *test-promote-rules
+  needs:
+    - job: test-promote-build
+      artifacts: true
+  before_script:
+    - apk add -q bash curl unzip coreutils
+    - docker info
+    - |
+      image_target_exports="$(bash .gitlab/ci/scripts/image_target_env.sh "${NVCM_VALUES_IMAGE_TARGET:-}")" || exit $?
```
ryanheffernan, line 99: We might want to protect against pipeline variables overriding this (assuming we don't want that)

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,365 @@
...
+      apt update
+      apt install -y helm git wget unzip
+      # Pinned yq + SHA-256 verification (this is a secret-bearing job; do not
+      # run an unverified binary from a floating 'latest' URL).
+      YQ_VERSION=v4.44.6
+      YQ_SHA256=0c2b24e645b57d8e7c0566d18643a6d4f5580feeea3878127354a46f2a1e4598
+      wget -qO /usr/local/bin/yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
+      echo "${YQ_SHA256}  /usr/local/bin/yq" | sha256sum -c -
+      chmod +x /usr/local/bin/yq
+    - wget --content-disposition https://api.ngc.nvidia.com/v2/resources/nvidia/ngc-apps/ngc_cli/versions/4.8.2/files/ngccli_linux.zip -O ngccli_linux.zip
```
ryanheffernan: Could we pin and verify the NGC CLI download as well?

.gitlab/ci/scripts/promote_push_images.sh
```diff
@@ -0,0 +1,98 @@
...
+
+    src="pr-local/${image}:${short_sha}"
+    dst="${NVCM_IMAGE_REPOSITORY}/${image}:${PROMOTE_VERSION}"
+    docker tag "$src" "$dst"
+    echo "Pushing ${dst}..."
+    docker push "$dst"
+
+    # Authoritative digest: ask the registry for the pushed manifest digest
+    # (the local image id is NOT the registry manifest digest).
+    digest="$(docker buildx imagetools inspect "$dst" --format '{{json .Manifest}}' | grep -o '"digest": *"sha256:[0-9a-f]*"' | head -n 1 | cut -d'"' -f4)"
```
ryanheffernan: i don't think the fallback below will ever run, If imagetools inspect fails or grep finds no digest, this assignment exits the script immediately

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,365 @@
...
+        echo "Recent deploy-states on ${NVCM_ENV_BRANCH} (newest first):"
+        echo ""
+        git log -n 15 --pretty='%h  %ad  %s' --date=iso -- "$state_file"
+        echo ""
+        echo "Re-run this pipeline with NVCM_ROLLBACK_TO=<commit sha> to roll back."
+        exit 1
+      fi
+
+      echo "Rolling ${NVCM_PROMOTE_ENV} back to deploy-state @ ${NVCM_ROLLBACK_TO}"
+      git show "${NVCM_ROLLBACK_TO}:${state_file}" > "$state_file"
```
ryanheffernan, line 461: we should verify that NVCM_ROLLBACK_TO belongs to the right branch, otherwise a typo or SHA from another branch could install an unrelated deploy-state

.gitlab-ci.yml
```diff
@@ -10,3 +10,4 @@ include:
   - local: .gitlab/ci/release.yml
   - local: .gitlab/ci/airgapped.yml
   - local: .gitlab/ci/pr-build.yml
+  - local: .gitlab/ci/promote-test-envs.yml
```
ryanheffernan, line 13: With this both old manual deployment buttons appear alongside the replacement workflow, we might want to remove or disable the legacy ones to prevent confusion

.gitlab/ci/promote-test-envs.yml
jojung1: You're right. And there's more to it cause pipeline variables could also override `NVCM_CHART_REPO`, `NVCM_VALUES_REPO_PATH` and `NVCM_TEST_ENV_TARGETS`, so whoever starts a promote could redirect where images are pushed or where deploy-state is committed.

That said, we can't disable overrides outright because the flow needs `NVCM_PROMOTE_PR`/`NVCM_PROMOTE_ENV` at runtime. Claude recommended having the minimum role required to override pipeline variables set to Maintainer (in repository CI/CD settings) because this would match the design intent that promotes are run by authorized users, and it bounds who could tamper with the rest. This makes sense to me but I wanted to hear your thoughts?

.gitlab-ci.yml
jojung1: Yea I was thinking we could disable the legacy test01 deploy button here once the new test01 appset gets merged in the argocd repo and we can confirm that it works.

jojung1: I am also leaving the deploy-to-test pipeline alone for now because it's being used for QA atm.

.gitlab/ci/promote-test-envs.yml
jojung1: Pinned to 4.8.2 with sha256sum -c before unzip. Downloaded the artifact and confirmed it contains ngc-cli/ngc. NGC doesn't publish a checksum alongside the download, so this is computed + pinned by us and it will need updating whenever we bump the version.

.gitlab/ci/scripts/promote_push_images.sh
jojung1: Added `|| true` to the primary lookup (and the fallback) so a failed imagetools inspect now falls through as intended.

.gitlab/ci/promote-test-envs.yml
jojung1: Added guards: 

1. SHA must resolve to a real commit
2. Must be an ancestor of the env branch
3. Must actually contain that env's deploy-state.yaml

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,417 @@
...
+      # to the env branch, so the overrides can move between this gate and the
+      # deploy-state commit. write_deploy_state refuses to write if they have.
+      env_branch_revision="$(git -C /tmp/values-repo rev-parse "origin/${NVCM_ENV_BRANCH}")"
+      echo "ENV_BRANCH_REVISION=${env_branch_revision}" >> "${CI_PROJECT_DIR}/chart.env"
+      # Attest the chart version too, now that it has been cross-checked against
+      # the built artifact's filename above. chart.env is published as a FILE
+      # artifact (not only dotenv) because pipeline variables outrank dotenv in
+      # GitLab - an operator-supplied variable of the same name would otherwise
+      # override the values these provenance guards compare against.
+      echo "PROMOTE_VERSION=${PROMOTE_VERSION}" >> "${CI_PROJECT_DIR}/chart.env"
```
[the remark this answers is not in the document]
jojung1: Already addressed in [761fa80](https://github.com/NVIDIA/nv-config-manager/pull/143/commits/761fa80770f24f77a08af265f276ae4275561f80). write_deploy_state.sh no longer reads any deployment-affecting value from dotenv — PROMOTE_VERSION, BASELINE_REVISION, ENV_BRANCH_REVISION come from chart.env and all nine digests from digests.env, both consumed as file artifacts via an attest() helper that fails on a missing key. Digests stay in digests.env deliberately so they're attested by the job that computed them rather than re-copied by a later job. Verified a hostile same-named environment variable no longer wins.

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,417 @@
...
+      # Download the chart .tgz packaged in the SECRET-FREE build stage on the
+      # PR ref (job id resolved from GitLab-trusted metadata by
+      # promote_build_pipeline.sh). This job never checks out or packages PR
+      # source, so no untrusted chart code (e.g. `helm dependency build`) runs
+      # in this credentialed context.
+      : "${CHART_BUILD_JOB_ID:?missing dotenv from test-promote-build}"
+      curl -fsS --retry 3 --retry-delay 5 -H "JOB-TOKEN: ${CI_JOB_TOKEN}" \
+        -o chart-artifacts.zip \
+        "${CI_API_V4_URL}/projects/${CI_PROJECT_ID}/jobs/${CHART_BUILD_JOB_ID}/artifacts"
+      unzip -o -q chart-artifacts.zip
```
ryanheffernan: we might want to extract into an isolated directory and make sure no key files are getting overritten by the extract (like test_env_config.sh) before going ahead

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,417 @@
...
+  image: $NVCM_CI_IMAGE_ALPINE
+  rules: *test-promote-rules
+  timeout: 2h
+  before_script:
+    - apk add --no-cache bash curl git jq coreutils
+  script:
+    - bash .gitlab/ci/scripts/promote_build_pipeline.sh
+  artifacts:
+    reports:
+      dotenv: promote.env
```
ryanheffernan, line 64: More of an optional hardening nit than a blocker but we could publish promote.env as an artifact and have consumers read it from disk which would prevent any accidental variable overriding. not sure if worth doing so just a comment

.gitlab/ci/scripts/test_env_config.sh
```diff
@@ -0,0 +1,71 @@
...
+  return 0
+}
+
+shell_export() {
+  printf 'export %s=%q\n' "$1" "$2"
+  return 0
+}
+
+requested_env="$(trim "${1:?usage: test_env_config.sh <env>}")"
+records="${NVCM_TEST_ENV_TARGETS:-}"
```
ryanheffernan, line 55: Worth adding a guard here that we aren't promoting to a prod or other unexpected environment?

.gitlab/ci/promote-test-envs.yml
ryanheffernan: yeah that's a good point, this isn't a huge risk since we control the pipeline variables. one option would be to use pipeline inputs   for the PR number and environment inputs, constrain the environment to test/test01, and retrict pipeline varables which would separate runtime inputs from env variables, but i don't think that should be a blocker for v1

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,417 @@
...
+    allow_failure: true
+  - when: never
+
+# -----------------------------------------------------------------------------
+# 1) Resolve PR HEAD; reuse or trigger the no-secrets build pipeline; wait.
+# -----------------------------------------------------------------------------
+test-promote-build:
+  stage: deploy
+  tags:
+    - dgxc-corp-restricted
```
ryanheffernan, line 51: can the restricted runners handle the promotion? do they need to talk to argo or the test envs?

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,417 @@
...
+        --set "global.images.temporalUi.digest=${DIGEST_NV_CONFIG_MANAGER_TEMPORAL_UI}" \
+        > /dev/null
+      echo "✅ Render OK"
+
+      echo "📤 Publishing chart to configured targets..."
+      # publish_chart_targets.sh resolves ngc-cli/ngc relative to the project
+      # root, so publish from there.
+      cp "$chart_file" "${CI_PROJECT_DIR}/${CHART_NAME}-${PROMOTE_VERSION}.tgz"
+      cd "${CI_PROJECT_DIR}"
+      bash .gitlab/ci/scripts/publish_chart_targets.sh "${CHART_NAME}-${PROMOTE_VERSION}.tgz" "$PROMOTE_VERSION" "$CHART_NAME"
```
ryanheffernan, line 297: Could we verify that an already-existing chart version matches the package this run rendered? Repeated or concurrent builds use the same PR/SHA version, and the publishing helper treats already exists as success. That could leave deploy-state pointing at an earlier package that this run didn’t actually validate.

.gitlab/ci/promote-test-envs.yml
```diff
@@ -0,0 +1,417 @@
...
+      # env-branch overrides + digest pins) BEFORE publishing or touching the
+      # env branch. helm template is sandboxed (no env access), so rendering the
+      # untrusted chart with the values token present cannot leak secrets; the
+      # packaged .tgz already vendors its subcharts, so nothing is fetched.
+      env_config="$(bash "${CI_PROJECT_DIR}/.gitlab/ci/scripts/test_env_config.sh" "$NVCM_PROMOTE_ENV")" || exit $?
+      eval "$env_config"
+      if [ -n "${NV_CONFIG_MANAGER_VALUES_REPO_URL:-}" ]; then
+        values_repo_url="$NV_CONFIG_MANAGER_VALUES_REPO_URL"
+      else
+        values_repo_url="https://oauth2:${NV_CONFIG_MANAGER_VALUES_PUSH_TOKEN}@${CI_SERVER_HOST}/${NVCM_VALUES_REPO_PATH}.git"
```
ryanheffernan: do we need the push token here? it looks like we're just reading so ideally we could use a RO token

.gitlab/ci/promote-test-envs.yml
jojung1: I think they should assuming DCA allowed egress to NGC and kiwi-argocd? They shouldn't need to talk to argocd or the test envs.

.gitlab/ci/promote-test-envs.yml
jojung1: Good call. Updated to use `CI_JOB_TOKEN` now.

ryanheffernan (approved):

PR #126, opened 2026-07-14
Network policy respects namespace from .Values.monitoring.

spidercensus:
## Description

External namespaces must be specifically provided access to the NVCM namespace via network policy.  Namespaces can be specified to be added to this policy whitelist at several points in the values file.  One point that was configurable but being ignored was .Values.monitoring.namespace.

This PR adds support for this configured namespace to be added to the network policy `namespace-isolation`.
## Validation

- [ ] Standard CI passes.
- [x] Kind integration passes, or this PR explains why it was not run.

First Kind integration test: https://github.com/NVIDIA/nv-config-manager/actions/runs/29350165629/job/87144062277

Latest kind integration test https://github.com/NVIDIA/nv-config-manager/actions/runs/29518761542
```text
/ok to test <sha>
/kind test
```

As a fallback, run Actions -> Kind Integration -> Run workflow against the copy-pr-bot generated
`pull-request/<PR_NUMBER>` branch. Use the default `test_path` for the full suite, or narrow it
only while debugging.

The completed workflow updates this PR description with its conclusion and exact run URL.

Passing Kind Integration run, if not automatically reported:

Kind integration completed with **success** for [`a7904f103471`](https://github.com/NVIDIA/nv-config-manager/actions/runs/29518761542).

## Checklist

- [x] I am familiar with the contributing guidelines in `CONTRIBUTING.md`.
- [x] Commits are signed off for DCO compliance.
- [x] New or existing tests cover these changes, or the PR explains why tests are not needed.
- [x] Documentation is updated for user-facing behavior changes.
- [ ] Generated artifacts are updated when applicable, such as OpenAPI specs,
      docs screenshots, or Helm/rendered outputs.

## Summary by CodeRabbit

* **New Features**
  * Added configurable **Prometheus namespace** for network-policy-controlled scraping.
  * Updated the Infrastructure/Monitoring UI with a **Prometheus namespace (network policy)** field (with validation).
  * When the bundled observability stack is enabled, Prometheus automatically uses the release namespace.
* **Bug Fixes**
  * Network policy now allows Prometheus scraping only when monitoring is enabled and the Prometheus namespace is configured, limited to the expected ports.
* **Documentation**
  * Updated observability/metrics guidance and installer mapping for `monitoring.prometheus.namespace` (default `monitoring`).
* **Tests**
  * Added tests covering defaulting, custom namespace overrides, and observability-enabled behavior.

PR #159, opened 2026-07-21
feat(helm): optional prometheus-nats-exporter sidecar + PodMonitor for nautobot NATS

### LINKED ISSUE #163: CI: extend the Kind Docker Hub containerd mirror to all deploy paths (currently security-only)
## Background

PR #122 added a Docker Hub pull-through **containerd mirror** for the Kind cluster to fix repeated HTTP 429 (anonymous Docker Hub rate-limit) failures on the security integration runs. See `scripts/configure-kind-dockerhub-mirror` and the `security: true` branch of `.github/workflows/kind-integration.yml`.

How it works:
- `nv-gha-runners/setup-proxy-cache` already configures the **runner host** with a Docker Hub mirror (writes `/etc/buildkit/buildkitd.toml`).
- Our script reads that endpoint and re-applies it to the **Kind node's containerd** via `containerdConfigPatches`, so in-cluster `docker.io` pulls are served from the cache (`https://dc.local.gha-runners.nvidia.com`) instead of hitting `registry-1.docker.io` anonymously.

It is currently **scoped to the security run only**, because that path pulls ~2x more `docker.io` images (SPIRE, Keycloak, cert-manager, Envoy, redis, nats, busybox, curl…) and was the only path actually hitting the limit.

## Proposal

Extend the mirror to **all** Kind deploy paths (default `make kind-up` + observability), as defense-in-depth. Docker Hub's anonymous limit (~100 pulls / 6h) is **per-IP and shared across all runs on the fleet**, so even the plain deploy's docker.io pulls (cnpg/postgres, redis, nats, app third-party deps) can get caught in the crossfire as PR concurrency grows.

### Suggested implementation
- Hoist mirror discovery to run **once** at the top of the `Deploy to Kind` step (before the `if/elif/else`): compute `KIND_CONFIG` (patched copy when a mirror is found, else the default) and reuse it in all three `make` calls.
- Keep the behavioral asymmetry:
  - **security**: mirror is required — `exit 1` if not discoverable (already the case; removed the image-preload fallback in #122).
  - **default / observability**: mirror is **best-effort** — if no mirror is discoverable (e.g. local dev / off-fleet), fall back to the plain `deploy/kind-config.yaml` and proceed (these paths survive anonymous pulls). Never fatal.
- The script is already a no-op off-fleet, so local `make kind-up` is unaffected.

### Validation
- Requires **one plain (default, non-security) run** to confirm the regular path works with the mirror — none of the current runs cover it (they're all `security: true`).

## Not doing now
Keeping #122 scoped to the security path to keep the change focused; this is the low-risk follow-up.

dmathur0:
## Summary

Adds an **opt-in** Prometheus metrics path for the in-cluster nautobot NATS, so JetStream consumer backlog is scrapeable — the prerequisite for KEDA render autoscaling on render-queue depth.

- **`prometheus-nats-exporter` sidecar** on the nautobot NATS deployment, behind `nautobotNats.metrics.enabled` (default `false`). Runs with `-jsz=all -connz -varz -prefix=nats`, polling the NATS monitoring endpoint (`:8222`) and exposing `nats_*` metrics on `:7777` (`prom-metrics` port).
- **NATS `PodMonitor`** gated on `monitoring.podMonitors.enabled AND nautobotNats.metrics.enabled`, selecting the NATS pod and scraping the `prom-metrics` port.
- **`global.images.natsExporter`** → `docker.io/natsio/prometheus-nats-exporter:0.15.0` (same version already proven in production).

The key flag is `-jsz=all`, which emits `nats_consumer_num_pending` — the series a future `renderService.autoscaling` KEDA trigger scales on.

## Why gate the PodMonitor on `metrics.enabled`

Rather than a standalone `podMonitors.nats.enabled` toggle, the PodMonitor is gated on `nautobotNats.metrics.enabled` (in addition to `monitoring.podMonitors.enabled`) so you can never render a PodMonitor pointing at a `prom-metrics` port that doesn't exist.

## Blast radius

None by default. Both knobs default to `false`; `helm template` with default values renders **no** sidecar and **no** PodMonitor — existing environments are byte-identical.

## Test plan

- [x] `helm template … --values values-ci.yaml --set nautobotNats.metrics.enabled=true` → sidecar (`natsio/prometheus-nats-exporter:0.15.0`, `-jsz=all`, `prom-metrics`) + PodMonitor render with correct selector/port.
- [x] `helm template … --values values-ci.yaml` (defaults) → zero exporter/PodMonitor resources.
- [ ] Deploy via a values overlay (`nautobotNats.metrics.enabled=true`, `monitoring.podMonitors.enabled=true`) and confirm Prometheus scrapes `nats_consumer_num_pending` with the expected labels.

_Note: enablement in a values overlay (+ `renderService.autoscaling`) and the render scale-test are follow-ups; this PR is the reusable chart plumbing only._

## CI: Docker Hub mirror for all Kind deploy paths

The exporter sidecar is a new **in-cluster** `docker.io` pull, which 429'd against the
shared runner IP's anonymous Docker Hub budget and stalled `helm --wait` into the 20m
deploy timeout. The containerd pull-through mirror added in #122 was scoped to the
`security` branch only, so the observability path this PR runs on had no cover.

Mirror discovery is now hoisted above the deploy branches and the patched Kind config is
handed to all three (`default`, `observability`, `security`). An undiscoverable mirror
stays **fatal** for security runs (no preload fallback there) and degrades to a
**warning** for the plain/observability runs, so off-fleet runners still get CI.

Closes #163.

## Kind Integration

Kind integration completed with **success** for [`5f03323`](https://github.com/NVIDIA/nv-config-manager/commit/5f03323865c28ebd4da784e8888128fffdf235f4) in [this workflow run](https://github.com/NVIDIA/nv-config-manager/actions/runs/30709951450).

## Summary by CodeRabbit

- **New Features**
  - Added optional Prometheus NATS metrics export for local Nautobot, including configurable port **7777**, JetStream monitoring, and automatic PodMonitor discovery.
  - Improved KEDA render autoscaling with in-cluster Prometheus discovery and scale-to-zero support.
  - Added KEDA installation and air-gapped image support.
- **Bug Fixes**
  - Corrected NATS stream configuration and Prometheus endpoint handling.
  - Improved autoscaling behavior when minimum replicas are set to zero.
- **Tests**
  - Expanded observability, autoscaling, installer, and air-gapped deployment coverage.
- **Chores**
  - Updated Kind CI to consistently use Docker Hub pull mirrors when available.


deploy/helm/values.yaml
```diff
@@ -139,6 +139,11 @@ global:
     natsBox:
       repository: docker.io/natsio/nats-box
       tag: "0.14.3"
+    # prometheus-nats-exporter sidecar (optional; nautobotNats.metrics.enabled).
+    # Polls the NATS monitoring endpoint and exposes nats_* Prometheus metrics.
```
ryanheffernan, line 143: this doesn't enable the exporter for air-gapped image installs, i assume we want it enabled for those?

deploy/helm/templates/monitoring.yaml
```diff
@@ -696,5 +697,45 @@ spec:
         {{- include "nv-config-manager.podMonitorMetricRelabelings" . | nindent 8 }}
       {{- end }}
 {{- end }}
+
+{{- /* Nautobot NATS exporter PodMonitor (JetStream metrics for KEDA autoscaling) */}}
+{{- if and .Values.monitoring.podMonitors.enabled .Values.nautobotNats.metrics.enabled }}
```
ryanheffernan: we should also prevent exporter installation when we use external nautobot/nats

deploy/helm/values.yaml
daveshm: fixed

deploy/helm/templates/monitoring.yaml
daveshm: fixed

deploy/helm/values.yaml
```diff
@@ -139,6 +139,11 @@ global:
...
+    natsExporter:
+      repository: docker.io/natsio/prometheus-nats-exporter
+      tag: "0.20.1"
```
ryanheffernan, line 146: The air-gapped install still references Docker Hub because natsExporter is not registered with the installer’s image mapping. Please add docker.io/natsio/prometheus-nats-exporter:0.20.1 to both _GLOBAL_IMAGE_DEFAULTS and IMAGE_OVERRIDE_KEYS, then extend the air-gap Helm-values test to verify it is rewritten to the configured registry.

deploy/helm/templates/nautobot.yaml
```diff
@@ -109,6 +109,29 @@ spec:
...
+            - -port={{ .Values.nautobotNats.metrics.port | default 7777 }}
+            - -connz
+            - -varz
+            - -jsz=all
+            - -prefix=nats
+            - -use_internal_server_id
+            - http://localhost:8222/
+          ports:
+            - containerPort: {{ .Values.nautobotNats.metrics.port | default 7777 }}
+              name: prom-metrics
```
ryanheffernan, line 133: We also need to allow this port in
[`deploy/helm/templates/network-policy.yaml`](https://github.com/NVIDIA/nv-config-manager/blob/d6ee8e1b4a6e35a71ce4ae42a603293338210083/deploy/helm/templates/network-policy.yaml#L37-L56);
otherwise Prometheus running in the configured monitoring namespace cannot scrape the exporter.

deploy/helm/values-observability.yaml
```diff
@@ -79,6 +79,23 @@ monitoring:
...
+# observability path (local-dev + the opt-in kind integration run) actually
+# exercises NATS JetStream metric collection — the prerequisite for KEDA render
+# autoscaling on nats_consumer_num_pending. The installer flips
+# monitoring.{enabled,podMonitors.enabled} on when observability_enabled; this
+# makes the sidecar + PodMonitor render alongside them (see
+# templates/nautobot.yaml and templates/monitoring.yaml). Both default to off in
+# values.yaml, so only this overlay (and real deployments that opt in) run them.
+# test_nats_exporter.py asserts the PodMonitor CR + exporter sidecar exist when
+# OBSERVABILITY=true.
+nautobotNats:
```
ryanheffernan, line 95: shoudl we should update the kind observability run to include the exporter as well?

deploy/helm/templates/nautobot.yaml
```diff
@@ -109,6 +109,29 @@ spec:
...
+        # and re-exposes nats_* Prometheus metrics on :{{ .Values.nautobotNats.metrics.port | default 7777 }}.
+        # -jsz=all emits the JetStream nats_consumer_num_pending series KEDA
+        # render autoscaling scrapes; -prefix=nats yields the nats_ metric names.
+        - name: prom-exporter
+          image: "{{ .Values.global.images.natsExporter.repository }}:{{ .Values.global.images.natsExporter.tag }}"
+          imagePullPolicy: {{ .Values.global.imagePullPolicy | default "IfNotPresent" }}
+          {{- include "nv-config-manager.containerSecurityContext" . | nindent 10 }}
+          args:
+            - -port={{ .Values.nautobotNats.metrics.port | default 7777 }}
+            - -connz
```
ryanheffernan: KEDA only requires -jsz=all. Plain -connz emits aggregate—not per-connection—metrics, but it still makes NATS enumerate connections on every scrape - do we plan to use these extra stats for dashboards?

deploy/helm/values-observability.yaml
dmathur0: It already is — the observability kind run picks up the exporter without a workflow change.

The `kind-integration.yml` observability path doesn't `helm -f values-observability.yaml` directly; it sets `infrastructure.monitoring.observability_enabled=true` in `ci-integration-test.yaml` and runs `make kind-up`. The installer's deployer then layers this file in via a second `-f`:

```python
# deployer.py
if self.config.infrastructure.monitoring.observability_enabled:
    observability_values = Path(chart) / "values-observability.yaml"
    ...
    helm_args.extend(["-f", str(observability_values)])
```

and `helm_values.py` flips `monitoring.podMonitors.enabled=true` on observability. So with `nautobotNats.metrics.enabled: true` set here, both the exporter sidecar and the NATS PodMonitor render on the observability run.

Confirmed empirically: the observability kind run (`observability=true`) passed `test_nats_exporter.py`, which is a hard failure (not a skip) when `OBSERVABILITY=true` and asserts the PodMonitor CR + exporter sidecar exist.

deploy/helm/templates/nautobot.yaml
dmathur0: Good catch — and it became a real gap after #126 merged, which turned the monitoring-namespace rule into an explicit TCP port allow-list (8000/8003/9000/9090/stork/9187). The exporter serves on `nautobotNats.metrics.port` (default 7777), which wasn't in that list, so Prometheus in a separate monitoring namespace would be dropped.

Fixed in bec2afc: added the exporter port to the allow-list, gated on `nautobotNats.metrics.enabled` so it only opens when the sidecar is deployed:

```yaml
        - protocol: TCP
          port: 9187  # CNPG metrics exporter
        {{- if .Values.nautobotNats.metrics.enabled }}
        - protocol: TCP
          port: {{ .Values.nautobotNats.metrics.port | default 7777 }}  # nautobot NATS exporter
        {{- end }}
```

deploy/helm/templates/nautobot.yaml
dmathur0: Agreed — we only use `-jsz=all` for KEDA (`nats_consumer_num_pending`), and there are no connection-level dashboards planned. Dropped `-connz` in 1561071, keeping `-varz` (cheap server-health) and `-jsz=all`, and updated the sidecar contract test's required args to match. This avoids the per-scrape connection enumeration you flagged.

ryanheffernan (dismissed):

deploy/helm/templates/_helpers.tpl
```diff
@@ -1045,6 +1045,28 @@ Result depends on configuration:
...
+a Prometheus KEDA never talks to.
+
+The FQDN is required because KEDA resolves it from its own namespace, not the
+release's. The service name comes from prometheus.server.fullnameOverride in
+values-observability.yaml; only the namespace is unknown to a values file, which
+is why this is resolved here.
+*/}}
+{{- define "nv-config-manager.inClusterPrometheusAddress" -}}
+{{- if and .Values.prometheus.enabled (not .Values.renderService.autoscaling.prometheus.serverAddress) -}}
+{{- printf "http://prometheus-server.%s.svc.cluster.local:9090" .Release.Namespace -}}
```
ryanheffernan: Could this derive the service name from prometheus.server.fullnameOverride?

deploy/helm/templates/render-service-scaledobject.yaml
ryanheffernan, line 102: Should we require natsMetricsNamespace when an external serverAddress is configured?

dmathur0:
## Review status and integration evidence

### Open threads

**@ryanheffernan** — four of your five are already in the branch; I think the threads just never got resolved:

| Point | Where it landed |
|---|---|
| Air-gap: register `natsExporter` with the installer image mapping | `helm_values.py:171` (`_GLOBAL_IMAGE_DEFAULTS`) and `schema.py:794` (override keys), with an air-gap test asserting the registry rewrite |
| Allow the exporter port in the NetworkPolicy | `network-policy.yaml:63`, gated on the configured monitoring namespace |
| Include the exporter in the kind observability run | `values-observability.yaml:95` sets `nautobotNats.metrics.enabled: true`; the run exercises it (see below) |
| Don't install the exporter for external nautobot/NATS | `monitoring.yaml:719` gates the PodMonitor on `externalServices.nautobot.local` |

On **`-connz`** — agreed, and it's already dropped. The sidecar runs `-varz -jsz=all -prefix=nats -use_internal_server_id` only, and `nautobot.yaml:117-119` records why: KEDA doesn't need it, and it makes NATS enumerate every connection per scrape for stats we don't dashboard.

**@coderabbitai** — valid, fixed in 14b6187. With autoscaling on, no in-cluster Prometheus, and no external `serverAddress`, the ScaledObject rendered `serverAddress: ""`; KEDA accepts that and then silently never scales.

I guarded the value where it's consumed rather than where it's resolved, so the error fires only when a ScaledObject is actually emitted — enabling autoscaling with no consumers still renders nothing, as before. The message names the consumer and both remedies:

```
Error: execution error at (render-service-scaledobject.yaml:79:26):
renderService.autoscaling is enabled for the device consumer but no Prometheus
endpoint is available. Set renderService.autoscaling.prometheus.serverAddress
to an external Prometheus, or enable the in-cluster one via prometheus.enabled.
```

I checked every deployment that renders a ScaledObject before applying this — all of them set an explicit external `serverAddress`, so no existing output changes and no sync can start failing on it.

### Integration run

Kind Integration passes with observability enabled: [run 30566310148](https://github.com/NVIDIA/nv-config-manager/actions/runs/30566310148) — 50 passed, 62 skipped, every step green.

All 15 new tests ran and passed, both `device` and `nautobot` variants:

```
test_keda_autoscaling.py::test_scaledobjects_exist_for_each_configured_consumer  PASSED
test_keda_autoscaling.py::test_trigger_prometheus_address_is_resolved[...]       PASSED
test_keda_autoscaling.py::test_trigger_query_filters_the_exporter_series[...]    PASSED
test_keda_autoscaling.py::test_scaledobject_becomes_ready[...]                   PASSED
test_keda_autoscaling.py::test_keda_serves_the_trigger_metric[...]               PASSED
test_keda_autoscaling.py::test_trigger_query_matches_a_live_series[...]          PASSED
test_nats_exporter.py::test_nats_deployment_has_exporter_sidecar                 PASSED
test_nats_exporter.py::test_nats_podmonitor_scrapes_exporter_port                PASSED
test_nats_exporter.py::test_nats_podmonitor_selects_nats_pod                     PASSED
test_nats_exporter.py::test_nats_podmonitor_excludes_nats_box                    PASSED
```

Note this workflow is `workflow_dispatch` against the mirror branch, so it doesn't attach to the PR as a check — the link above is the only place it shows up.

That run covers f533ac1. A re-run for the guard commit is in flight: [run 30572611458](https://github.com/NVIDIA/nv-config-manager/actions/runs/30572611458).

deploy/helm/values.yaml
dmathur0: resolved and added a test for this

deploy/helm/templates/_helpers.tpl
dmathur0: Yes, and it turned out to be a live bug rather than just a cleanup — fixed in bd60b36.

The helper hardcoded `prometheus-server` while the comment three lines above claimed the name came from `prometheus.server.fullnameOverride`. Those agreed only because `values-observability.yaml:157` happens to set that exact string. Override it and both the ScaledObject's `serverAddress` and the NetworkPolicy rule that opens KEDA's path would point at a service that doesn't exist — and since they share this helper, they'd be consistently wrong together, so nothing would flag it.

Now derived, with the upstream chart's naming as the fallback when no override is set:

```gotemplate
{{- $server := .Values.prometheus.server | default dict -}}
{{- $name := $server.fullnameOverride | default (printf "%s-prometheus-server" .Release.Name) -}}
{{- printf "http://%s.%s.svc.cluster.local:9090" $name .Release.Namespace -}}
```

Verified the override is actually tracked now:

```
--set prometheus.server.fullnameOverride=custom-prom
  serverAddress: "http://custom-prom.nv-config-manager-monitoring.svc.cluster.local:9090"
```

Default output is unchanged, and all existing deployments render byte-identical.

deploy/helm/templates/render-service-scaledobject.yaml
daveshm: https://github.com/NVIDIA/nv-config-manager/pull/159/changes/bd60b36bbf6d0a7f841d78302fe6409aeff2d50a#diff-f2a3fa403688b4d3c925b3b9aa992e4bada3e1390c3068b51da782074e0a652dR38 it's already here no?

daveshm: i think i committed that after your comment :P

dmathur0:
## Kind Integration: green on current head

[Run 30591723295](https://github.com/NVIDIA/nv-config-manager/actions/runs/30591723295) passed on `5bf0380e571c7be82c88ecd8c34430a03f5c71ca` in 22 minutes.

```
50 passed, 62 skipped, 1 warning in 55.25s
```

That SHA is both the current PR head and the tip of the `pull-request/159` mirror the workflow builds from, so the green result covers the code exactly as it stands. Nothing is in flight.

The run was dispatched with `observability: true`, which is what exercises this PR — the default `/kind test` path hardcodes observability off, so the exporter and KEDA resources are never deployed there and these tests would skip. That is also why this run does not appear in the PR's check list: `workflow_dispatch` runs are not associated with the PR.

All 15 tests covering this change ran rather than skipped:

**`test_nats_exporter.py`**
- `test_nats_deployment_has_exporter_sidecar`
- `test_nats_podmonitor_selects_nats_pod`
- `test_nats_podmonitor_scrapes_exporter_port`
- `test_nats_podmonitor_excludes_nats_box`

**`test_keda_autoscaling.py`** (parametrized over the `device` and `nautobot` consumers)
- `test_scaledobjects_exist_for_each_configured_consumer`
- `test_scaledobject_becomes_ready`
- `test_trigger_prometheus_address_is_resolved`
- `test_trigger_query_filters_the_exporter_series`
- `test_trigger_query_matches_a_live_series`
- `test_keda_serves_the_trigger_metric`

`test_trigger_query_matches_a_live_series` is the one that matters most: it replays the exact PromQL the ScaledObject renders against the live in-cluster Prometheus and asserts a series comes back. That closes the failure mode this PR exists to prevent, where a trigger is accepted by KEDA, reports `READY=True`, and silently never scales because its query matches nothing.

The earlier 503 on `test_diagnostics_ticketless.py` was a transient workflow-service flake unrelated to these changes and did not recur.

dmathur0:
## Default-path run, and a correction

Ran Kind Integration with `observability: false` to confirm this PR doesn't disturb the default deployment path, since everything it adds is gated behind observability.

**The gating works.** [Run 30656652662](https://github.com/NVIDIA/nv-config-manager/actions/runs/30656652662) on `5bf0380e` reported `33 passed, 78 skipped`, and all 15 exporter and KEDA tests skipped cleanly rather than erroring — 78 skipped here against 62 in the observability run is exactly those 15 plus the parametrized cases. No exporter sidecar, no PodMonitor, no ScaledObjects are rendered when observability is off, which is the intent.

**Correcting my earlier comment.** I wrote above that the `test_diagnostics_ticketless.py::test_ticketless_with_tech_support` 503 "did not recur." That was wrong — it recurred on this run, with the same error:

```
requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url:
  https://workflow.config-manager.local/v1/workflow/2fb3dd84-...
```

Checking the last ten Kind Integration runs, every failure is this one test, and nothing else. So it is a standing flake rather than a one-off, and it is unrelated to this PR — it fails identically with observability on and off, and touches none of the code here.

The mechanism is that `_poll_to_terminal` calls `resp.raise_for_status()` on every poll, so any single transient 503 from the workflow service aborts the test outright instead of being retried like a `ConnectionError` is. Worth fixing separately; I can file it if useful.

## Also in this push

`a3ecf32` corrects a comment I had written on the `natsMetricsNamespace` guard. It claimed the namespace label is what keeps NATS series distinct across environments on a shared Prometheus. That isn't true: cells reuse namespace names, and I confirmed two unrelated cells reporting under the same namespace into the same Mimir. What actually separates environments is `consumer_name`, which carries `nats.queue`.

The guard itself is unchanged and still worth having — it narrows the query so that a `consumer_name` which stops being unique doesn't scale a deployment on another environment's backlog — but the stated reasoning would have misled the next reader. Comment-only; rendered output is identical, and I verified the guard still fires with an external `serverAddress` and no namespace, and still renders with one.

A fresh observability run is in flight on `a3ecf32`.

dmathur0:
## Green again on `a3ecf32`

[Run 30660067523](https://github.com/NVIDIA/nv-config-manager/actions/runs/30660067523) passed on `a3ecf32a9ec817af2f1b3b1efeb285c3feeb7ebd` in 22 minutes, with `observability: true`.

```
50 passed, 62 skipped, 1 warning in 53.98s
```

All 15 exporter and KEDA tests passed rather than skipped, matching the earlier run on `5bf0380e` exactly — as expected, since `a3ecf32` is comment-only. That SHA is the current PR head and the mirror tip, so the result covers the code as it stands, and the standing ticketless 503 did not appear this time.

Both deployment paths are now verified on this PR:

| | observability on | observability off |
|---|---|---|
| run | [30660067523](https://github.com/NVIDIA/nv-config-manager/actions/runs/30660067523) on `a3ecf32` | [30656652662](https://github.com/NVIDIA/nv-config-manager/actions/runs/30656652662) on `5bf0380e` |
| exporter + KEDA tests | 15 passed | 15 skipped cleanly |
| result | 50 passed, 62 skipped | 33 passed, 78 skipped, 1 pre-existing flake |

All 20 PR checks are green and the branch is mergeable, pending review.

dmathur0:
## Both paths green on the rebased head

Re-ran Kind Integration after the merge with main. All three runs on `5f03323865c28ebd4da784e8888128fffdf235f4` passed, covering both deployment configurations:

| run | observability | result | exporter + KEDA tests |
|---|---|---|---|
| [30674422578](https://github.com/NVIDIA/nv-config-manager/actions/runs/30674422578) | true | 50 passed, 62 skipped, 0 failed | 15 passed |
| [30703712220](https://github.com/NVIDIA/nv-config-manager/actions/runs/30703712220) | false | 34 passed, 78 skipped, 0 failed | 15 skipped |
| [30709951450](https://github.com/NVIDIA/nv-config-manager/actions/runs/30709951450) | false | 34 passed, 78 skipped, 0 failed | 15 skipped |

`5f03323` is the current PR head and the mirror tip, so this covers the code as it stands. The incoming changes from main touch only `temporal/ngc/activities/backup.py` and its tests, nothing near the exporter or KEDA paths, and the results match the pre-merge runs exactly.

With observability on, all 15 exporter and KEDA tests execute and pass. With it off, all 15 skip cleanly and none of the sidecar, PodMonitor, or ScaledObject resources are rendered — the feature is fully gated.

### On the earlier red default-path run

The failure I reported previously on `5bf0380e` was `test_diagnostics_ticketless.py::test_ticketless_with_tech_support` hitting a 503, and it is unrelated to this PR. Three consecutive default-path runs have since passed it, and the same failure has occurred on #193, so it is repo-wide rather than branch-specific.

Cause, for whoever picks it up: `_poll_to_terminal` calls `resp.raise_for_status()`, which raises `HTTPError` on a 503, but the retry clause catches only `ConnectionError`. A single transient 503 anywhere in the 600-second poll window therefore aborts the test instead of being retried — in the failing run it polled once, saw `status=RUNNING`, then died on the next tick with 597 seconds still on the clock. The helper's own docstring says transient failures are logged and retried, so this is a gap rather than a deliberate choice. The fix is to retry on 5xx while still failing fast on 4xx. I'll raise it separately rather than widening this PR.

deploy/helm/templates/kubernetes-secrets.yaml
```diff
@@ -126,6 +126,29 @@ spec:
               creds_path = {{ .Values.externalServices.nats.credsPath }}
               {{- end }}
               local = {{ .Values.externalServices.nats.local }}
+              {{- /*
+              Stream and subject names, mirroring the vault-secrets.yaml [nats]
```
zsblevins, line 130: I would drop everything but the first sentence, AI really loves putting a bunch of irrelevant historical context in comments that just leads to bloat :)

installer/src/nv_config_manager_installer/helm_values.py
```diff
@@ -634,6 +635,24 @@ def _build_postgres_section(pg: ExternalPostgresConfig) -> dict[str, Any]:
     return postgres
 
 
+# The chart defaults these to the legacy "kiwi" names that the ESO deployments
```
zsblevins: Drop the comment here, we can also remove this, I'm going to file another PR to fix the kiwi names that wormed their way back into this repo, we should never have kiwi references in this repo

deploy/helm/templates/kubernetes-secrets.yaml
dmathur0: Done in 981bf7e — trimmed to just the first sentence.

Fair hit, and the same instinct was in a few other comments on this PR. I've left the rest alone for now to keep this diff reviewable, but happy to do a pass if you spot others.

installer/src/nv_config_manager_installer/helm_values.py
dmathur0: Comment dropped in 981bf7e.

On removing the constant itself — agreed that it should go, but it can't come out on its own without relocating a live stream, so I've left it for your PR where the deletion becomes a no-op.

The reason: the chart default for this stream is `kiwi`. Before this PR the kubernetes-secrets path omitted the stream keys from the assembled INI entirely, so the app fell back to its own default, `nv-config-manager`, and that is what those deployments are running on today. This PR closes that gap and starts delivering the keys — which means without the pin they would render `kiwi` and the running install would move to a different stream, orphaning its durables and stranding whatever is pending.

Once the chart default is `nv-config-manager`, the pin is redundant and deleting it changes nothing. That is exactly your PR, so it is the right place for it.

One thing worth knowing going in: the ESO deployments genuinely are on `kiwi` today, since they take it from the chart default. I checked kiwi-argocd and none of the environment values files pin the stream name — they all inherit it. So flipping the chart default moves their streams too, and they would need `kiwi` pinned explicitly first, or a drain and cut-over per environment. Happy to help with that sequencing if useful.

zsblevins (approved):

PR #168, opened 2026-07-23
fix(dhcp): reconcile KEA config using native config hash

dmathur0:
## Description

Fixes the DHCP recovery failure observed during EKS upgrades. When the `kea` container restarts independently of the `config-sync-v4` sidecar, Kea comes back up on the bootstrap `/etc/kea/kea-dhcp4.conf` (no subnets, default memfile lease DB). Because the desired config in Redis is unchanged, the old sync loop — which only compared the Redis value against its in-process `previous_config` — concluded "nothing changed" and never re-applied to Kea, leaving the pod unhealthy indefinitely.

This switches drift detection to Kea's native SHA-256 effective-config hash (Kea 2.4+), so any loss or modification of the running config is detected and reconciled regardless of whether Redis changed.

- `KeaClient.set_config()` now returns the hash from the `config-set` response (`str | None`; `None` on pre-2.4 Kea).
- New `KeaClient.get_config_hash()` using `config-hash-get`.
- `_sync_kea_configuration_async` retains an `expected_hash`:
  - **Startup:** unconditionally applies the desired Redis config and records the verified hash (restart-safe).
  - **Redis changed:** apply new config, replace `expected_hash`.
  - **Redis unchanged:** compare `config-hash-get` to `expected_hash`; on mismatch (e.g. Kea restarted to bootstrap) re-apply the desired config and update `expected_hash`.
- `_apply_and_verify_kea_config` verifies the effective hash after applying. A `KeaException` or `TimeoutError` on `config-hash-get` after a successful `config-set` is treated as an unverified apply (`None`) so the sidecar does not crash-loop over a config that is already applied.
- Avoids comparing full `config-get` JSON (which contains Kea-generated defaults).

Part 1 of the DHCP config-sync recovery series. Heartbeat liveness (#171), strict traffic gating (#170), and observability (#172) stack on this.

## Validation

- `uv run pytest src/tests/dhcp/test_cli.py` — hash reconcile tests, including startup `KeaException` and `TimeoutError` on `config-hash-get`.
- Local kind: reproduced the Kea-container-restart incident on `main` (0 subnets, `memfile`, pod still Ready, sidecar silent), deployed this branch, and confirmed the sidecar logged a hash mismatch and reapplied within one refresh interval without container restarts.

- [x] Standard CI passes.

The GitHub Kind Integration workflow was not dispatched for this change. The DHCP recovery path was validated on a local kind cluster against a real Kea 2.6 / Redis / CNPG deploy (see the kind write-up on this PR). Full-suite kind CI is not required to exercise this sidecar loop.

Kind integration completed with **success** for [`9974b3d`](https://github.com/NVIDIA/nv-config-manager/commit/9974b3d67e35ebac6f2590cf6fdeeb6a15077ef8) in [this workflow run](https://github.com/NVIDIA/nv-config-manager/actions/runs/31738884557).

## Checklist

## Summary by CodeRabbit

- **Reliability Improvements**
  - KEA synchronization now verifies that applied settings match the running configuration.
  - Automatically detects configuration drift and KEA restarts, then reapplies settings when needed.
  - Configuration updates now expose KEA’s effective configuration hash.
  - Improved handling of synchronization and verification failures, allowing later retries where appropriate.

- **Tests**
  - Added coverage for configuration updates, drift recovery, restarts, unchanged configurations, and failure scenarios.

### DIFF
--- a/src/nv_config_manager/dhcp/cli.py
+++ b/src/nv_config_manager/dhcp/cli.py
@@ -225,10 +225,11 @@ async def _apply_and_verify_kea_config(
     applied_hash = await kea_client.set_config(config, version=ip_version)
     try:
         effective_hash = await kea_client.get_config_hash(version=ip_version)
-    except KeaException as exc:
+    except (KeaException, TimeoutError) as exc:
         # config-set already succeeded, so the desired config is applied and
-        # persisted -- only the verification read failed. The refresh loop
-        # tolerates this same failure, and aborting here would instead crash-loop
+        # persisted -- only the verification read failed. get_config_hash
+        # re-raises TimeoutError (it does not wrap it in KeaException), and the
+        # refresh loop already swallows both. Aborting here would crash-loop
         # the sidecar over a config that is actually applied. Returning None
         # leaves the next drift check to reapply and re-verify.
         logger.warning(f"Could not verify the applied KEA configuration hash: {exc}")
--- a/src/nv_config_manager/dhcp/cli.py
+++ b/src/nv_config_manager/dhcp/cli.py
@@ -26,7 +26,7 @@ import click
 
 from nv_config_manager.common.config import load_config
 from nv_config_manager.common.log import LogCategory, configure_logging, get_logger
-from nv_config_manager.dhcp.kea import KeaClient
+from nv_config_manager.dhcp.kea import KeaClient, KeaException
 from nv_config_manager.dhcp.kea_dhcp_confgen import generate_config, inject_lease_db_config
 from nv_config_manager.dhcp.metrics import DHCP_CACHE_REFRESH_ERRORS
 from nv_config_manager.dhcp.nautobot import NautobotClient
@@ -208,6 +208,31 @@ def refresh_kea_configuration(
     asyncio.run(_refresh_loop_async(ip_version, check, refresh_interval))
 
 
+async def _apply_and_verify_kea_config(
+    kea_client: KeaClient,
+    config: dict[str, Any],
+    ip_version: int,
+) -> str | None:
+    """Apply the desired configuration to KEA and return its verified hash.
+
+    KEA (2.4+) returns the SHA-256 hash of the effective configuration from
+    ``config-set``. Before treating the sync as successful, the running
+    configuration is confirmed against that hash via ``config-hash-get`` so a
+    configuration that was rolled back or only partially applied surfaces as an
+    error rather than being silently trusted. The verified effective hash is
+    returned to track for subsequent drift detection.
+    """
+    applied_hash = await kea_client.set_config(config, version=ip_version)
+    effective_hash = await kea_client.get_config_hash(version=ip_version)
+    if applied_hash is not None and effective_hash != applied_hash:
+        raise KeaException(
+            f"KEA effective configuration hash ({effective_hash}) does not match "
+            f"the hash returned by config-set ({applied_hash}); "
+            "the configuration was not applied cleanly."
+        )
+    return effective_hash
+
+
 async def _sync_kea_configuration_async(
     ip_version: int,
     refresh_interval: int,
@@ -232,9 +257,10 @@ async def _sync_kea_configuration_async(
         # so that secrets are not stored in the Redis cache
         config = inject_lease_db_config(config, ip_version)
 
-        # Run once
+        # Run once. The startup path always applies the desired Redis config and
+        # captures a fresh effective hash, which keeps a config-sync restart safe.
         logger.info(f"Setting initial KEA DHCPv{ip_version} Configuration from Redis.")
-        await kea_client.set_config(config, version=ip_version)
+        expected_hash = await _apply_and_verify_kea_config(kea_client, config, ip_version)
 
         if refresh_interval:
             logger.info(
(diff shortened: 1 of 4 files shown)


dmathur0:
## Validated end-to-end on a local kind cluster

Reproduced the original incident against a real deployment (Kea 2.6, CNPG Postgres lease DB, Redis-backed config), confirmed the pre-change behaviour, then deployed this branch and re-ran the identical experiment.

### Why the Kea restart reverts the config

`/etc/kea` is not a volume mount — it lives in the Kea container's writable layer. When kubelet restarts that container, the layer is recreated from the image and `/etc/kea/kea-dhcp4.conf` reverts to the bootstrap file, discarding whatever `config-write` had persisted. So a Kea container restart, independent of the sidecar, is sufficient to reproduce the incident.

### Baseline (`main`)

Verified the running sidecar had zero `get_config_hash` references first, then killed supervisord in the Kea container only:

| | before | after Kea restart |
|---|---|---|
| `subnet4` entries | 1 | **0** |
| `lease-database.type` | `postgresql` | **`memfile`** |
| `config-hash-get` | `F2E5D76A5DD7B835C0FE…` | `3CF47F35DD057E88CCF9…` |
| pod status | `4/4` Ready | `4/4` **Ready** |

The pod stayed Ready while unable to serve a single lease, so it remained a valid external DHCP target. The sidecar logged nothing: Redis still held the desired config, it matched the sidecar's own in-memory copy, and with `debug=False` the unchanged-config path is silent. It was behaving exactly as written and still left Kea unconfigured, because nothing compared against what Kea was *actually running*.

### Recovery on `main` is not absent, but it is slow and violent

Worth recording, because it reframes what this PR buys. The config mismatch fails the shared `/healthcheck`, which trips several liveness probes in turn (kubelet events):

```
07:45:43  healthcheck      Liveness probe failed: statuscode 500
07:45:51  config-sync-v4   Liveness probe failed: statuscode 500   <- #65 backstop starts counting
07:46:29  api              Container api failed liveness probe, will be restarted
07:46:33  healthcheck      Container healthcheck failed liveness probe, will be restarted
07:48:38  kea              Container kea failed liveness probe, will be restarted
```

Note the Kea recycle at 07:48:38 — that restart puts Kea straight back onto the bootstrap config, so it does not help.

The backstop `livenessProbe` added in #65 (`httpGet /healthcheck:9090`, `periodSeconds: 30`, `failureThreshold: 10`) began failing at 07:45:51 and would have recycled `config-sync-v4` at roughly 07:50:21, healing via the unconditional "Run once" `set_config()` at startup. **I did not observe it fire**: the rollout of this branch deleted the pod at 07:49:47, about 34 seconds early. So on `main` the drift does self-heal, in ~5 minutes, after recycling three other containers.

### With this branch

Same kill, against the same deployment:

```
07:51:46  kea container exits (Completed/0)
07:51:47  kea container restarts on the bootstrap config
07:51:53  config-sync-v4 detects the drift and reapplies
```

```
WARNING KEA running configuration hash
(3CF47F35DD057E88CCF9849C66F8E7242F97A9EEAABAFBA91C12C43724F8D8B8) does not match the expected
hash (F2E5D76A5DD7B835C0FE880E713CAB9226CFCAA85084959C83141FB05E78BC6E); reapplying desired
configuration (KEA may have restarted).
```

Those are the same two hashes recorded in the baseline run, so the drift and the recovery are both confirmed against known values. Kea returned to 1 subnet on `postgresql` at hash `F2E5D76A…`, with zero `Error refreshing the KEA config` entries.

Container restart counts on the test pod afterwards:

```
api              restarts=0
config-sync-v4   restarts=0
healthcheck      restarts=0
kea              restarts=1   <- the injected fault
```

**Net effect: recovery goes from a ~5 minute, four-container recycle cascade to a 7 second in-place reapply with nothing recycled.**

### Two incidental findings

`config-hash-get` returned `result: 0` against the deployed server, empirically confirming hash support is present — `build/kea.Dockerfile` installs from the `kea-2-6` ISC repo, so the pre-2.4 fallback discussed in review is not reachable for any pod built from this repo.

**Merge ordering with #171.** Because the #65 backstop probes the shared `/healthcheck`, it currently doubles as config-drift recovery rather than only wedge recovery — that is what the timeline above shows. #171 replaces it with the self-contained `check-sync-heartbeat` exec probe, which by design does *not* fail on a config mismatch. That is correct once this PR handles drift in-loop, but if #171 lands first there is a window where neither the in-loop check nor the probe recovers bootstrap drift. Suggest landing this PR first, or landing the two together.

src/nv_config_manager/dhcp/cli.py
```diff
@@ -208,6 +208,40 @@ def refresh_kea_configuration(
...
+    ``config-set``. Before treating the sync as successful, the running
+    configuration is confirmed against that hash via ``config-hash-get`` so a
+    configuration that was rolled back or only partially applied surfaces as an
+    error rather than being silently trusted. The verified effective hash is
+    returned to track for subsequent drift detection.
+    """
+    applied_hash = await kea_client.set_config(config, version=ip_version)
+    try:
+        effective_hash = await kea_client.get_config_hash(version=ip_version)
+    except KeaException as exc:
```
ryanheffernan: should we also catch `TimeoutError` here? otherwise a timeout on getting the hash will escape

ryanheffernan (commented):
also can you please update the pr description to match the repo template?

src/nv_config_manager/dhcp/cli.py
dmathur0: Yes — `get_config_hash` re-raises `TimeoutError` instead of wrapping it in `KeaException`, so a hung `config-hash-get` after a successful `config-set` was still aborting startup.

Caught `(KeaException, TimeoutError)` in `_apply_and_verify_kea_config`, same unverified-apply path (`return None`). The refresh loop already swallowed both via `except Exception`. Added `TimeoutError` to the startup regression (`test_startup_hash_get_failure_does_not_abort_sync`).

ryanheffernan (approved):
also can you please update the pr description to match the repo template?

PR #172, opened 2026-07-23
feat(dhcp): add reconcile observability metrics and structured logs

dmathur0:
## Description

Adds enough state to diagnose a future DHCP upgrade failure without relying on container restart counts alone — the signal that was missing during the incident. Instruments the sync and refresh loops with three metrics and a structured log-state taxonomy, with strict secret redaction.

### Metrics

| Name | Type | Labels |
|------|------|--------|
| `nv_config_manager_dhcp_sync_failures_total` | Counter | `operation`, `ip_version` |
| `nv_config_manager_dhcp_config_hash_mismatches_total` | Counter | `ip_version` |
| `nv_config_manager_dhcp_last_successful_sync_timestamp_seconds` | Gauge | `ip_version` |

`operation` is one of `redis_read`, `config_generation`, `config_set`, `hash_get`, `config_test`, `postgres`, so operators can tell *where* reconciliation broke. The gauge is a Unix timestamp; alert on its age (`time() - gauge`).

Both counters and the gauge are seeded before the first reconcile attempt. `prometheus_client` only materializes a labeled child on first use, so a pod whose initial sync never succeeds would otherwise export no series at all — and an alert on the age of a missing series cannot fire.

### Scrape path

These counters live in process-local registries, and neither confgen command serves other HTTP traffic, so each one exports its own registry on `CONFGEN_METRICS_PORT` (9091). The `/metrics` endpoint in `dhcp/api.py` is a different process and does not include them.

- `config-sync-v4`: `sync-metrics` containerPort, scraped by the `<dhcp>` PodMonitor.
- `config-refresh-v4`: `refresh-metrics` containerPort and its own `<dhcp>-refresh` PodMonitor, because it is a separate Deployment that the `app: <dhcp>` selector does not match.
- Both bind the same port (separate pods, so no conflict) and NetworkPolicy already allows 9091 from the monitoring namespace.
- `refresh-kea-configuration --check` skips the exporter: it is a one-shot validation run that exits before any scrape.

A bind failure is logged and swallowed rather than raised, so a metrics problem can never stop reconciliation.

### Structured logs (`sync_state` field)

`waiting-for-initial-redis-config`, `in-sync`, `drift-detected` (logs `desired_hash` and `running_hash`), `applying` (logs `desired_hash`), `dependency-error` (logs `operation` and the error).

Redaction is strict: only truncated SHA-256 hashes are logged, never credentials or full `config-get` responses, and structured field values are newline-escaped to prevent log forging. That includes the unhandled-exception path — tracked operations sanitize before counting but re-raise, so the `sys.excepthook` handler redacts as well and attaches only stack frames, which carry no exception message.

### Failures that do not raise

Two paths reported a failure without an exception, so the tracked-operation wrapper counted them as successes:

- `KeaClient.test_config` returns `(False, error)` on a validation rejection. Now counted explicitly. This also fixes an adjacent ordering bug inherited from `main`, where `--check` returned before the validity branch and reported an invalid config as valid.
- A `config-set` hash that disagrees with the immediately retrieved effective hash. Both calls succeed, so the raise bypassed the wrappers; the mismatch counter and a `config_set` failure are now recorded before raising.

### Rebase note

`cli.py` overlaps with #168 (merged), #170, and #171. Whichever of the remaining two merges later needs a small rebase of `_sync_kea_configuration_async`.

## Validation

- `uv run pytest src/tests/dhcp/` — 130 passed, 1 pre-existing Docker-only failure (`test_kea_config_valid`).
- `uv run ruff check` / `ruff format --check` and `uv run mypy` on the touched files — clean.
- `helm template test . --values values-ci.yaml --values values-observability.yaml` renders both PodMonitors; verified the `<dhcp>-refresh` selector matches the rendered refresh pod labels.
- `test_confgen_metrics_port_is_scrapeable` asserts the port constant, both named containerPorts, both PodMonitor endpoints, and the NetworkPolicy entry stay consistent.

- [x] Standard CI passes.

Kind integration passed on `0057c46`, recorded below. The commit after it adds the refresh-side exporter and PodMonitor, so a re-run is worthwhile before merge even though the change is covered by the render and unit tests above.

Kind integration completed with **success** for [`100c141`](https://github.com/dsx-ai-factory/nv-config-manager/commit/100c1416183da2f9eceaa075482988216b9371bd) in [this workflow run](https://github.com/dsx-ai-factory/nv-config-manager/actions/runs/35027512299).

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for DHCP synchronization, refresh activity, failures, configuration drift, and successful sync times.
  * Added heartbeat-based liveness checks and improved DHCP refresh connectivity.
  * Added monitoring for DHCP refresh processes and reconciliation metrics.
  * Added support for scraping configuration reconciliation metrics.

* **Bug Fixes**
  * Improved timeout, retry, validation, and configuration hash verification handling.
  * Improved drift detection and recovery, including behavior during service outages.
  * Improved startup behavior when configuration digests are unavailable.
  * Improved health checks to distinguish process liveness from configuration readiness.


src/nv_config_manager/dhcp/cli.py
```diff
@@ -117,15 +264,21 @@ async def _refresh_kea_configuration_async(
...
+            redis_client=redis_client,
+            version=ip_version,
+            kea_config=current_kea_config,
+        ),
     )
 
     logger.info("Validating configuration against KEA API.")
-    result, error = await kea_client.test_config(config, version=ip_version)
+    result, error = await _track_sync_operation(
+        SyncOperation.CONFIG_TEST, ip_version, kea_client.test_config(config, version=ip_version)
```
ryanheffernan, line 340: CONFIG_GENERATION and CONFIG_TEST are recorded by config-refresh-v4, but only config-sync-v4 starts a metrics server, and since the registry is process-local these failure metrics are never scraped.

src/nv_config_manager/dhcp/cli.py
```diff
@@ -16,25 +16,172 @@
...
+    operation: str,
+    ip_version: int,
+    awaitable: Awaitable[T],
+) -> T:
+    """Await ``awaitable``, recording a labeled failure before re-raising."""
+    try:
+        return await awaitable
+    except Exception as exc:
+        _record_sync_failure(operation, ip_version, exc)
+        raise
```
ryanheffernan, line 179: re-raising after sanitizing the log could result in the original unsanitized message being logged by the exception handler which could leak creds

src/nv_config_manager/dhcp/metrics.py
```diff
@@ -40,3 +81,31 @@
...
+
+# Unix timestamp (seconds) of the last successful, verified synchronization of
+# the desired config onto the local Kea server. Operators alert on the AGE of
+# this value: ``time() - <gauge>`` climbing means reconciliation has stalled,
+# which is exactly the signal that was missing during the EKS-upgrade incident.
+DHCP_LAST_SUCCESSFUL_SYNC_TIMESTAMP = Gauge(
+    "nv_config_manager_dhcp_last_successful_sync_timestamp_seconds",
+    "Unix timestamp of the last successful verified DHCP synchronization",
+    ["ip_version"],
+)
```
ryanheffernan, line 137: we should initialize the gauge to zero (or error/alert on an absent series) otherwise there will be no series if initial sycnhronization fails

src/nv_config_manager/dhcp/cli.py
```diff
@@ -117,15 +264,21 @@ async def _refresh_kea_configuration_async(
...
+            nautobot_client=nautobot_client,
+            redis_client=redis_client,
+            version=ip_version,
+            kea_config=current_kea_config,
+        ),
     )
 
     logger.info("Validating configuration against KEA API.")
-    result, error = await kea_client.test_config(config, version=ip_version)
+    result, error = await _track_sync_operation(
```
ryanheffernan, line 339: KeaClient.test_config reports validation rejection as (False, error), so _track_sync_operation sees a successful await and does not increment config_test

src/nv_config_manager/dhcp/cli.py
```diff
@@ -232,7 +387,11 @@ async def _apply_and_verify_kea_config(
...
         # the sidecar over a config that is actually applied. Returning None
         # leaves the next drift check to reapply and re-verify.
-        logger.warning(f"Could not verify the applied KEA configuration hash: {exc}")
+        _record_sync_failure(SyncOperation.HASH_GET, ip_version, exc)
+        logger.warning(
+            "Could not verify the applied KEA configuration hash: %s",
+            _safe_error_text(exc),
+        )
         return None
     if applied_hash is not None and effective_hash != applied_hash:
```
ryanheffernan, line 473: A mismatch between the config-set hash and the immediately retrieved effective hash raises outside the tracked operations and does not increment the mismatch counter

src/nv_config_manager/dhcp/cli.py
dmathur0: You're right, and it's worse than a scrape-config gap: `config-refresh-v4` is its own Deployment (`<dhcp>-refresh`), so it had no container port, no PodMonitor, and the existing DHCP PodMonitor's `app: <dhcp>` selector never matched it. Those two labels were unreachable by construction.

Fixed in 91b815b8 by giving that process an exporter rather than dropping the labels:
- `refresh-kea-configuration` now starts the exporter and seeds its own series.
- `refresh-metrics` containerPort on the refresh container, plus a `<dhcp>-refresh` PodMonitor whose selector matches its pod labels.
- Same port for both processes (separate pods, so no conflict), so the constant is now `CONFGEN_METRICS_PORT` instead of `SYNC_METRICS_PORT`. The NetworkPolicy already allowed 9091 namespace-wide.
- Skipped under `--check`, which is a one-shot validation run that exits before any scrape could happen.

`test_confgen_metrics_port_is_scrapeable` now asserts both named ports and both PodMonitor endpoints, and I confirmed the refresh PodMonitor's selector matches the rendered pod labels via `helm template` with the observability overlay.

src/nv_config_manager/dhcp/cli.py
dmathur0: Confirmed, and I traced where it landed: `_exception_handler`, installed as `sys.excepthook` in non-debug mode. It logged the exception raw *and* passed `exc_info`, which appends the original message again. So a Redis/PostgreSQL DSN redacted by `_record_sync_failure` went straight back into the log once the re-raise escaped `asyncio.run`. The startup Redis read and the whole refresh loop (which deliberately lets errors kill the container) both take that path.

Fixed in 91b815b8: the handler now logs `_safe_error_text(exception)` and attaches only the formatted stack frames, which carry no exception message, run through the same redaction. `exc_info` is gone for that reason. Test asserts the password is absent and that frames are still present for diagnosis.

src/nv_config_manager/dhcp/metrics.py
dmathur0: Agreed, and this defeated the main alert: `prometheus_client` only materializes a labeled child on first use, so a pod whose initial sync never succeeded exported no `last_successful_sync_timestamp` series at all, and an alert on the age of a missing series cannot fire. Exactly the case the gauge exists for.

Fixed in 91b815b8 with `initialize_sync_metrics`, called before the first reconcile: gauge seeded at 0 (so the pod reports an unbounded age), and the mismatch, cache-refresh, and per-operation failure counters materialized too, since `rate()` needs a prior sample.

Seeding is split per process on purpose. `initialize_refresh_metrics` seeds only `config_generation` / `config_test` and deliberately does not touch the gauge: that process never applies config to KEA, so a sync timestamp it can never advance would alert forever. Tests assert both directions, including that each process does not seed the other's labels.

src/nv_config_manager/dhcp/cli.py
dmathur0: Correct, `test_config` returns `(False, error)` instead of raising, so the wrapper saw a successful await and `config_test` never moved. Fixed in 91b815b8 by recording the failure explicitly on the falsy result. `_safe_error_text` now accepts a message as well as an exception, since KEA reports this one by return value.

While there I found an adjacent bug in the same branch and fixed it: `if check:` returned *before* `if not result:`, so `refresh-kea-configuration --check` logged "Generated configuration is valid." and exited 0 on a config KEA had just rejected. That ordering predates this PR (it's on `main`), so say the word if you'd rather it moved to its own change. Both behaviours are covered by tests.

src/nv_config_manager/dhcp/cli.py
dmathur0: Right, both calls return successfully there, so the raise never passes through a tracked wrapper and the mismatch counter stayed flat on what is arguably the most serious drift signal we have: KEA not keeping what `config-set` just reported. Fixed in 91b815b8 by incrementing the mismatch counter and recording a `config_set` failure before raising, with a comment noting why it can't rely on the wrappers. Test asserts both counters.

src/nv_config_manager/dhcp/cli.py
```diff
@@ -280,42 +530,100 @@ async def _sync_kea_configuration_async(
...
+                        desired_hash = _config_fingerprint(new_config)
+                        running_hash = expected_hash or "none"
+                        DHCP_CONFIG_HASH_MISMATCHES.labels(ip_version=str(ip_version)).inc()
+                        _log_sync_state(
+                            SyncState.DRIFT_DETECTED,
+                            "KEA DHCP configuration drift detected, updating.",
+                            ip_version,
+                            desired_hash=desired_hash,
+                            running_hash=running_hash,
+                        )
```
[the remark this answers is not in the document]
dmathur0: Valid, and the conflation was introduced by this PR — on `main` that branch only logged "Configuration changed, updating KEA DHCP Configuration." with no drift language and no counter. Fixed in 5fa07f0d.

`new_config != previous_config` compares two Redis snapshots, so it means config-generation published a new desired config. Kea's effective hash is never read on that path, so nothing is known to have diverged, and reporting the cached `expected_hash` as `running_hash` asserted a disagreement that was never observed.

I took the second option you offered (separate state) rather than reading `get_config_hash` first, because we are about to apply on this path regardless — an extra Kea round-trip plus a new `HASH_GET` failure mode on every config change buys nothing. The path now logs a new `desired-config-updated` state with `desired_hash` and `previous_desired_hash`, and no `running_hash`.

The counter is now incremented only in the `else` branch, which does read the hash back and compare. That matters operationally: the whole point of the metric is alerting, and the old behaviour would have fired drift alerts on every routine config push.

Tests: `test_sync_loop_increments_mismatch_and_logs_hashes` encoded the old semantics and is replaced by `test_desired_config_update_is_not_counted_as_drift` (asserts the counter does not move, `DRIFT_DETECTED` is absent, and no `running_hash` attribute). Since this left the verified-drift increment as the only one in the loop, I also pinned it in `test_kea_restart_bootstrap_recovery_with_unchanged_redis`, which now asserts the counter increments and that `running_hash` is Kea's actual `BOOTSTRAP_HASH` rather than a cached stand-in. Narrowed the metric help text and `SyncState` docstring to match.

src/nv_config_manager/dhcp/cli.py
```diff
@@ -16,25 +16,205 @@
...
+
+def _safe_error_text(error: BaseException | str) -> str:
+    """Return a log-safe, bounded error string with credentials stripped.
+
+    Accepts a message as well as an exception because KEA reports a validation
+    rejection as a returned error string rather than by raising.
+    """
+    raw = error if isinstance(error, str) else f"{type(error).__name__}: {error}"
+    text = _redact_secrets(escape_log_newlines(raw))
+    if len(text) > _MAX_ERROR_CHARS:
```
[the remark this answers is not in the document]
dmathur0: Both parts were valid and are fixed — the `ClickException` in c6e703cc, and the regex refined in 5fa07f0d.

The `ClickException` now passes the KEA error through `_safe_error_text`. Confirmed the leak first: a rejected config is echoed back by KEA and Click renders the exception to stderr unredacted, so the lease-DB password was reachable. Covered by `test_refresh_rejection_message_redacts_quoted_secret`, which asserts on both the CLI message and the failure log.

On the pattern, I first wrote the quoted-key support with a backreference so the quotes had to balance. Your version matches them independently, and testing the two side by side showed that difference is load-bearing:

| input | backreference | independent quotes |
| --- | --- | --- |
| `"password": "hunter2"` | redacted | redacted |
| `password": "hunter2"` | **leaks** | redacted |
| `password' : "hunter2"` | **leaks** | redacted |

Asymmetric quoting is exactly what malformed or partially-escaped KEA output can produce, and for a redaction helper over-redacting is harmless while under-redacting is the failure that matters. So I adopted the independent form. Both asymmetric cases are now regression-tested in `test_redact_secrets_covers_quoted_and_bare_keys`.

One behaviour change worth flagging: capturing key-plus-separator in group 1 preserves the original separator, so output is `"password": <redacted>` rather than being normalised to `password=<redacted>`. That is more faithful to the source text — it no longer invents an `=` where the JSON had a `:`. Test expectations updated accordingly.

src/nv_config_manager/dhcp/cli.py
```diff
@@ -280,42 +540,106 @@ async def _sync_kea_configuration_async(
...
+                            raise
+                        if kea_running_hash != expected_hash:
+                            DHCP_CONFIG_HASH_MISMATCHES.labels(ip_version=str(ip_version)).inc()
+                            _log_sync_state(
+                                SyncState.DRIFT_DETECTED,
+                                "KEA DHCP configuration drift detected, updating.",
+                                ip_version,
+                                desired_hash=expected_hash or "none",
+                                running_hash=kea_running_hash or "none",
+                            )
```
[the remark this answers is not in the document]
dmathur0: Fixed in 71208230.

The mismatch counter and `DRIFT_DETECTED` now fire only when `expected_hash` is known and disagrees with Kea's running hash. An unverified apply (`expected_hash is None` after a failed `config-hash-get`) still reapplies, but logs `applying` instead of drift. `test_startup_hash_get_failure_does_not_abort_sync` now asserts the counter stays flat.

Also caught `aiohttp.ClientError` on the post-`config-set` verification read so a transport blip is an unverified apply rather than a retry of a config Kea is already running.

dmathur0:
@coderabbitai please confirm the remaining review-body findings against HEAD (`71208230`):

1. Unknown `expected_hash` counted as confirmed drift — gated. The mismatch counter and `DRIFT_DETECTED` fire only when `expected_hash` is known and disagrees with Kea's running hash. An unverified apply still reapplies, but logs `applying`.
2. `aiohttp.ClientError` after `config-set` — caught alongside `KeaException`/`TimeoutError` in `_apply_and_verify_kea_config`, recorded as `HASH_GET`, and returned as `(None, False)` so a transport blip does not retry `config-set`.
3. The Sep 10 nitpick about an unbounded startup `except Exception` retry is stale after the #171 merge. Startup apply is bounded by `STARTUP_APPLY_ATTEMPTS` and then hands off to the monitoring loop; a rejected config no longer retries forever.

zsblevins (dismissed):

zsblevins (approved):

PR #180, opened 2026-07-24
fix(security): NATS account boundaries and JetStream API prefixes

dmathur0:
## Summary
- Cherry-picks prior work onto `main` so GitLab MR image builds can publish test images.
- Adds configurable JetStream API prefixes per stream so render consumers/admin APIs can reach account-scoped / imported streams (e.g. Nautobot via a custom `$JS.*.API` prefix).
- Wires the prefixes through Helm values, vault/kubernetes secret ini generation, config helpers, pull consumers, and admin consumer APIs.

## Test plan
- [ ] Confirm GitLab MR pipeline runs `docker-build-mr-amd64` / `docker-build-mr-arm64` / `docker-manifest-mr`
- [ ] Verify images tagged with `$CI_COMMIT_SHORT_SHA` land in the internal registry
- [ ] Unit tests for pull consumer + admin API prefix selection pass
- [ ] In a multi-account NATS env, set `externalServices.nats.streams.nautobot.apiPrefix` to the imported-account prefix and confirm the Nautobot consumer connects; config-manager streams remain on `$JS.API`

## Summary by CodeRabbit

* **New Features**
  * Added per-service JetStream API prefix configuration for config-manager, render-change, device-change, archive, and Nautobot.
  * Updated admin and pull consumer flows to use the configured JetStream prefix per consumer type.
  * Implemented consistent defaults and fallback behavior when prefixes aren’t set.
* **Documentation**
  * Updated Helm, Vault, and INI generation plus chart values to emit the new `*_api_prefix` settings.
* **Tests**
  * Expanded and added coverage to verify JetStream prefix propagation across admin consumer management, pull consumers, and archive wiring.

PR #184, opened 2026-07-27
fix(security): NATS account boundaries and JetStream API prefixes

dmathur0:
## Summary
- Cherry-picks prior work onto `main` so GitLab MR image builds can publish test images.
- Adds configurable JetStream API prefixes per stream so render consumers/admin APIs can reach account-scoped / imported streams (e.g. Nautobot via a custom `$JS.*.API` prefix).
- Wires the prefixes through Helm values, vault/kubernetes secret ini generation, config helpers, pull consumers, and admin consumer APIs.

Replaces #180 (same commits; branch renamed to drop an internal ticket key from the branch name).

## Notable fixes beyond the prefix plumbing

All four secret-rendering templates emit `config_manager_api_prefix` and `nautobot_api_prefix` and nothing else, so no deployment's stream or subject naming changes.

- **Publishers honor the prefix.** Both publish paths in `render/events/util.py` opened JetStream with a bare `nats_conn.jetstream()`, pinning them to `$JS.API` regardless of configuration.
- **`pull_subscribe` binds the stream explicitly.** Without `stream=`, the client resolves the stream via `$JS.API.STREAM.NAMES`, which is not exported across an account boundary and would block the Nautobot consumer outright.
- **Typed error handling at the boundary.** String matching on `"not found"` is replaced with `NotFoundError` / `ServiceUnavailableError`, so a transient error no longer looks like a missing consumer and triggers a spurious create attempt.
- **Reset fast-forwards instead of recreating.** See below.
- **Retry escalation.** Repeated consumer-cycle failures escalate from warning to error after a threshold rather than logging at the same level forever.

## Reset no longer deletes consumers

Reset used to delete the consumer and let the service recreate it with `DeliverPolicy.NEW`, which landed the cursor at the end of the stream. That depends on `CONSUMER.DELETE` and `CONSUMER.CREATE`. Neither is exported across the account boundary — the `nv-config-manager` account imports exactly three services for the nautobot stream:

```
$JS.API.CONSUMER.INFO.nautobot.>      -> $JS.CEREBRO.API.CONSUMER.INFO.nautobot.>
$JS.API.CONSUMER.MSG.NEXT.nautobot.>  -> $JS.CEREBRO.API.CONSUMER.MSG.NEXT.nautobot.>
$JS.ACK.nautobot.>                    (no remap)
```

There is no `STREAM.*` export either, so purge and `delete_msg` are equally unavailable. Fetching and acking is the only way to move a cursor on that stream.

Reset now binds to the existing durable and walks it past its backlog, acking without processing. One mechanism works on both locally owned and imported streams, `delete_consumer` is gone from the codebase entirely, and the 409 an earlier commit in this PR added is no longer needed.

Implementation notes:
- `pull_subscribe_bind` issues no JetStream API call and requires the stream to be named, so it also avoids the unexported `STREAM.NAMES` lookup that previously broke the consumer.
- The drain snapshots `num_pending` on entry, so a busy stream can't keep it looping forever.
- Batches stay under the consumer's `max_ack_pending` (live value is 1,000), and acks are flushed between batches since they're fire-and-forget publishes.
- A wall-clock budget bounds the request; a backlog too large for one pass returns `status: partial` with `skipped` and `remaining_pending` so the caller can re-invoke.

**Behavior changes for callers:** `ConsumerResetResponse` gains `skipped` and `remaining_pending`. Reset on a non-existent consumer now returns 404 rather than a synthetic success, since there is no cursor to advance. `status` can now be `partial`.

**What this does not cover:** deleting a consumer also reset its create-time configuration. Draining cannot change `max_deliver`, `ack_wait`, or `DeliverPolicy`, so consumer-config changes still have to land before the prefix flip or go through natsReady.

## Known follow-ups (not in this PR)

- Consumer config (`max_deliver`, `backoff`, `ack_wait`) is create-only — `_ensure_consumer_exists` never updates. Changing it requires delete + recreate, which stops being possible for the Nautobot consumer once the prefix flips, so those changes must land before the flip.
- Error-handling work in the dispatch paths (terminal vs. retryable classification) is tracked separately.

## Test plan
- [ ] Confirm GitLab MR pipeline runs `docker-build-mr-amd64` / `docker-build-mr-arm64` / `docker-manifest-mr`
- [ ] Verify images tagged with `$CI_COMMIT_SHORT_SHA` land in the internal registry
- [x] Unit tests for pull consumer + admin API prefix selection pass
- [x] Unit tests for publisher prefix selection on both `queue_render` and `queue_render_batch`
- [x] Unit tests for the drain: batch sizing, ack flushing, partial progress, and that reset never calls `delete_consumer`
- [ ] Against a real backlog, confirm reset drains it and that a concurrent running consumer doesn't break the drain
- [ ] In a multi-account NATS env, set `externalServices.nats.streams.nautobot.apiPrefix` to the imported-account prefix and confirm the Nautobot consumer connects; config-manager streams remain on `$JS.API`

## Summary by CodeRabbit

* **New Features**
  * Added configurable JetStream API prefixes and durable consumer names for Config Manager, Nautobot, and archive processing.
  * Added automatic provisioning and reconciliation of missing NATS consumers.
  * Added NATS consumer monitoring dashboards for pending, acknowledged, redelivered, and waiting messages.
  * Upgraded the NATS image to version 2.14.

* **Bug Fixes**
  * Improved handling of NATS permissions, missing consumers, and consumer resets with clearer operator guidance.

* **Documentation**
  * Documented external NATS deployments, required permissions, consumer settings, and new observability metrics.


src/nv_config_manager/temporal/client/nats.py
```diff
@@ -115,6 +128,8 @@ def __init__(
             subject=subject,
             queue_suffix=queue_suffix,
             handler=handler,
+            api_prefix=api_prefix
```
ryanheffernan: this could lead to a mismatch, e.g. `api_prefix` may target `archive_stream`, while `default_stream_name` remains `config_manager_stream`

PR #222, opened 2026-08-07
fix(build): preserve tracked build files during clean

ryanheffernan:
## Description

Update `make clean` to avoid deleting the `build/` directory.

## Validation

```
➜  nv-config-manager git:(rheffernan/preserve-root-build-on-clean) ls build
configure-apt-mirror.sh      kea-admin                    kea.Dockerfile               nautobot-test.Dockerfile     nv-config-manager.Dockerfile temporal.Dockerfile
kea                          kea-admin.Dockerfile         nats-ready.Dockerfile        nautobot.Dockerfile          setup.stork.deb.sh           ui.Dockerfile
➜  nv-config-manager git:(rheffernan/preserve-root-build-on-clean) make clean
find . -type d -name "__pycache__" -exec rm -rf {} + 2>/dev/null || true
find . -type d -name ".pytest_cache" -exec rm -rf {} + 2>/dev/null || true
find . -type d -name ".mypy_cache" -exec rm -rf {} + 2>/dev/null || true
find . -type d -name ".ruff_cache" -exec rm -rf {} + 2>/dev/null || true
find . -type d -name "*.egg-info" -exec rm -rf {} + 2>/dev/null || true
find . -type d -name "dist" -exec rm -rf {} + 2>/dev/null || true
find . -mindepth 2 -type d -name "build" -exec rm -rf {} + 2>/dev/null || true
find . -type d -name ".cache" -exec rm -rf {} + 2>/dev/null || true
➜  nv-config-manager git:(rheffernan/preserve-root-build-on-clean) ls build
configure-apt-mirror.sh      kea-admin                    kea.Dockerfile               nautobot-test.Dockerfile     nv-config-manager.Dockerfile temporal.Dockerfile
kea                          kea-admin.Dockerfile         nats-ready.Dockerfile        nautobot.Dockerfile          setup.stork.deb.sh           ui.Dockerfile
```

- [x] Standard CI passes.

Makefile change, no need to run kind

```text
/ok to test <sha>
```

```text
/kind test
```

only while debugging.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup behavior to remove only nested build directories, helping preserve top-level build directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,7 +386,7 @@ clean:
 	find . -type d -name ".ruff_cache" -exec rm -rf {} + 2>/dev/null || true
 	find . -type d -name "*.egg-info" -exec rm -rf {} + 2>/dev/null || true
 	find . -type d -name "dist" -exec rm -rf {} + 2>/dev/null || true
-	find . -type d -name "build" -exec rm -rf {} + 2>/dev/null || true
+	find . -mindepth 2 -type d -name "build" -exec rm -rf {} + 2>/dev/null || true
 	find . -type d -name ".cache" -exec rm -rf {} + 2>/dev/null || true
 
 # UI targets


Makefile
```diff
@@ -386,7 +386,7 @@ clean:
 	find . -type d -name ".ruff_cache" -exec rm -rf {} + 2>/dev/null || true
 	find . -type d -name "*.egg-info" -exec rm -rf {} + 2>/dev/null || true
 	find . -type d -name "dist" -exec rm -rf {} + 2>/dev/null || true
-	find . -type d -name "build" -exec rm -rf {} + 2>/dev/null || true
+	find . -mindepth 2 -type d -name "build" -exec rm -rf {} + 2>/dev/null || true
```
kxtran, line 389: Out of curiosity: What is the reason that you want to keep the Dockerfiles during clean? 

If you only want to preserve only the Dockerfiles, this would remove others: `find ./build -mindepth 1 -not -name "*.Dockerfile" -depth -delete`

ryanheffernan: this was deleting the checked-in dockerfiles we need to build the environment, so i had to `git restore` after running `make clean`. we also have build scripts in the `build/` directory we want to preserve so the proposed command would delete those as well. this command mistakenly cleared out the whole build dir

kxtran: got it. I was looking at the title, so I thought that we only wanted to keep the dockerfiles.

kxtran (approved):

Makefile
ryanheffernan: ah, good call. fixed the title to be accurate to what is changed

PR #226, opened 2026-08-10
Implement Pluggable DCIM

zsblevins:
## Validation

- [ ] Standard CI passes.

```text
/ok to test <sha>
```

```text
/kind test
```

only while debugging.

Kind integration completed with **success** for [`9513ca8`](https://github.com/dsx-ai-factory/nv-config-manager/commit/9513ca8d540423df130b2cab738d5f41b6921cb6) in [this workflow run](https://github.com/dsx-ai-factory/nv-config-manager/actions/runs/33018989946).

## Checklist


components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+                    raw_component, f"device '{device_name}' firmware component '{name}'"
+                )
+                components.append(
+                    RenderFirmwareComponent(
+                        name=str(name),
+                        artifact=RenderFirmwareArtifact(
+                            version=_optional_text(component.get("version")),
+                            image_file=_optional_text(component.get("file")),
+                            source_path=_optional_text(component.get("s3_path")),
+                        ),
```
kxtran: The same mapping appears 3 times in this func and it can be extracted into a helper function like:

```
def _firmware_artifact(value: Mapping[str, Any], image_key: str = "file") -> RenderFirmwareArtifact:
    return RenderFirmwareArtifact(
        version=_optional_text(value.get("version")),
        image_file=_optional_text(value.get(image_key)),
        source_path=_optional_text(value.get("s3_path")),
    )
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+    parent = value.get("parent")
+    return RenderLocation(
+        id=_optional_text(value.get("id")),
+        name=_required_text(value, "name", "location"),
+        kind=_named_text(location_type, "location type"),
+        tags=_tags(value.get("tags", ()), "location"),
+        parent=_location(_mapping(parent, "location parent"))
+        if isinstance(parent, Mapping)
+        else None,
+    )
```
kxtran, line 156: A circular parent chain could blow up the stack and I think that it would be good to have a depth guard

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+            esi_base_mac=_optional_text(context.get("evpn_esi_base_mac")),
+            fabric_mac=_optional_text(context.get("fabric-mac")),
+            df_preference=int(evpn_context["df-preference"])
+            if evpn_context is not None and evpn_context.get("df-preference") is not None
+            else None,
+        ),
+    )
+
+
+def _overlay_data(
```
kxtran: I think this func could benefit from extracting 3 helpers. This would let us unit-test L2/L3 VNI logic independently without wiring up a full overlay payload. Something like:

```
def _l2_vnis(raw_vxlans, ...) -> tuple[list[RenderL2Vni], dict[str, RenderL2Vni], dict[str, RenderL2Vni]]: ...
def _l3_vnis(raw_vxlans, ...) -> list[RenderL3Vni]: ...
def _l2_vni_vrfs(raw_assignments, l2_by_id, l2_by_overlay) -> list[RenderL2VniVrf]: ...
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+    value = record.get(key)
+    if value is None or value == "":
+        raise DCIMInvalidDataError(f"Nautobot {description} is missing required field '{key}'")
+    return str(value)
+
+
+def _optional_text(value: object) -> str | None:
+    """Normalize a nullable Nautobot scalar to text."""
+    if value is None or value == "":
+        return None
```
kxtran: I think this deserves its own helper that can be used everywhere the file checks for null-or-empty — `raw_vni not in (None, "")`, `vrf.get("name") not in (None, "")`, and also `_optional_text` and `_named_text` which spell it as `value is None or value == ""`. A single predicate keeps the idiom consistent and the call sites read more naturally (not _is_blank(...) vs the current double-negative not in (None, ""))
```
def _is_blank(value: object) -> bool:
    return value is None or value == ""
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+    )
+    return RenderData(
+        device=DeviceRenderData(
+            identity=RenderDeviceIdentity(
+                id=_required_text(device, "id", "device"),
+                name=device_name,
+                platform=_required_text(platform, "name", "device platform"),
+                role=_required_text(role, "name", "device role"),
+                model=_required_text(device_type, "model", "device type"),
+                location=_location(location),
```
kxtran: `_location(location)` is called twice (lines 1010 and 1025) with the same recursive walk. We can compute it once into a variable and pass that to both _location_data and RenderDeviceIdentity.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+    """Read a named Nautobot object without leaking its response shape."""
+    if value is None:
+        if required:
+            raise DCIMInvalidDataError(f"Nautobot {description} is missing required field 'name'")
+        return None
+    record = _mapping(value, description)
+    name = _optional_text(record.get("name"))
+    if name is None and required:
+        raise DCIMInvalidDataError(f"Nautobot {description} is missing required field 'name'")
+    return name
```
kxtran: `_named_text(..., required=True)` guarantees a non-None return (it raises otherwise), but the return type is still `str | None`, so every call site needs a redundant or "" guard. A simple fix: extract the required path into its own function that returns str, and the or "" calls can go away.
```
def _required_named_text(value: object, description: str) -> str:
    name = _named_text(value, description)
    if name is None:
        raise DCIMInvalidDataError(f"Nautobot {description} is missing required field 'name'")
    return name
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+
+def build_render_data(
+    device_response: Mapping[str, Any], location_response: Mapping[str, Any]
+) -> RenderData:
+    """Map Nautobot data and expose only SDK-level data-contract errors."""
+    try:
+        return _build_render_data(device_response, location_response)
+    except DCIMInvalidDataError:
+        raise
+    except (TypeError, ValueError, ValidationError) as exc:
```
kxtran, line 1114: Catching `TypeError` here risks masking programming errors (e.g. calling a method on None, wrong argument types) by wrapping them as data-contract errors. ValueError and ValidationError make sense — those come from bad data. But a TypeError usually means a bug in this module, not bad input from Nautobot. I'd drop TypeError from the catch so those surface as real crashes instead of getting silently wrapped.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
```diff
@@ -0,0 +1,1067 @@
...
+    addresses = _address_collection(
+        source_interface.get("ip_addresses", ()), f"device '{device_name}' BGP peer interface"
+    )
+    return RenderBGPPeer(
+        name=_required_text(peer_device, "name", f"device '{device_name}' BGP peer device"),
+        status=_named_text(
+            routing_instance.get("status"), f"device '{device_name}' BGP peer status", required=True
+        )
+        or "",
+        description=_required_text(peer_device, "name", f"device '{device_name}' BGP peer device"),
```
kxtran: The `name` and `description` are set the same value. Should description come from a different field?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/client.py
```diff
@@ -217,6 +191,9 @@ async def graphql_query(
         """
         session = await self._ensure_session()
         payload = {"query": query, "variables": variables or {}}
+        operation_name = getattr(query, "operation_name", None)
+        if operation_name is not None:
+            payload["operationName"] = operation_name
```
kxtran, line 201: the `query` is a str, when would it reach here?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+
+def _passes_ztp_aggregate_filter(
+    status: dict[str, Any] | None, is_aggregate_managed: bool | None
+) -> bool:
+    """Return whether a device passes ZTP and aggregate-management filters."""
+    if status and not status.get("ztp_enabled", True):
+        return False
+    return not (status and status.get("is_aggregate_managed", False) != is_aggregate_managed)
```
kxtran:
```suggestion
    if not status:
        return True
    if not status.get("ztp_enabled", True):
        return False
    return status.get("is_aggregate_managed", False) == is_aggregate_managed
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
```diff
@@ -0,0 +1,254 @@
...
+        "device_name": iface["device"]["name"],
+        "device_id": iface["device"]["id"],
+    }
+
+
+def _passes_ztp_aggregate_filter(
+    status: dict[str, Any] | None, is_aggregate_managed: bool | None
+) -> bool:
+    """Return whether a device passes ZTP and aggregate-management filters."""
+    if status and not status.get("ztp_enabled", True):
```
kxtran: Nit: double negation is hard to follow

The not (status and ... != ...) on the return line takes a moment to parse. An early-return style makes the logic more obvious in the below suggestion

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
```diff
@@ -0,0 +1,254 @@
...
+        return False
+    return not (status and status.get("is_aggregate_managed", False) != is_aggregate_managed)
+
+
+def _pool_ip_matches_prefix(pool_ip: dict[str, Any], prefix_entry_id: str, family: int) -> bool:
+    """Return whether a pool address belongs to the requested prefix and family."""
+    return bool(pool_ip["ip_version"] == family and pool_ip["parent"]["id"] == prefix_entry_id)
```
kxtran: Nit: bool() is a no-op here. The `==` already returns bool.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
```diff
@@ -0,0 +1,254 @@
...
+def _get_reservations_for_prefix(
+    all_reserved_ips: list[dict[str, Any]],
+    prefix_entry: dict[str, Any],
+    family: int,
+    is_aggregate_managed: bool | None,
+) -> list[dict[str, Any]]:
+    """Build provider-normalized reservations for a prefix."""
+    reservations: list[dict[str, Any]] = []
+    for reserved_ip in all_reserved_ips:
+        if reserved_ip["ip_version"] != family or reserved_ip["parent"]["id"] != prefix_entry["id"]:
```
kxtran: Nit: this duplicates the check in `_pool_ip_matches_prefix`. Could we rename that helper to something generic (e.g. `_ip_matches_prefix`) and reuse it here?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
```diff
@@ -0,0 +1,254 @@
...
+    """Build an eligible option candidate from a pool address."""
+    interfaces = pool_ip.get("interfaces")
+    if not interfaces or len(interfaces) > 1:
+        return None
+    iface = interfaces[0]
+    if not _passes_ztp_aggregate_filter(
+        iface["device"].get("configmanagerdevicestatus"), is_aggregate_managed
+    ):
+        return None
+    if not iface.get("mac_address") and not iface["device"].get("serial"):
```
kxtran: Nit: lines 100 and 154 share the same "must have MAC or serial" check — worth extracting into a small helper `like _has_identifier(iface)`.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def _get_pool_ips_and_candidates(
+    all_pool_ips: list[dict[str, Any]],
+    prefix_entry: dict[str, Any],
+    family: int,
+    is_aggregate_managed: bool | None,
+) -> tuple[list[Any], list[dict[str, Any]]]:
+    """Build pool addresses and option candidates for a prefix."""
+    pool_ips: list[Any] = []
```
kxtran: nit: 
```suggestion
    pool_ips: list[ipaddress.IPv4Address | ipaddress.IPv6Address] = []
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
```diff
@@ -0,0 +1,254 @@
...
+NV_CONFIG_MANAGER_MANAGED_VLANS = [
+    "vlan13",
+    "vlan30",
+    "vlan40",
+    "vlan997",
+    "vlan998",
+    "vlan1002",
+    "vlan1001",
+    "vlan901",
+]
```
kxtran: `NV_CONFIG_MANAGER_MANAGED_VLANS` doesn't appear to be referenced anywhere in the codebase. Is this still needed, or can we remove it?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/events.py
```diff
@@ -0,0 +1,352 @@
...
+)
+
+
+async def configcontext(
+    event: DCIMChangeEvent, client: DCIMClient
+) -> tuple[RenderEventRequest, ...]:
+    """Handle an ``extras.configcontext`` event."""
+    filters: dict[str, list[str]] = {}
+    for field in _CONFIG_CONTEXT_FILTER_FIELDS:
+        values = _record(event).get(field, ())
```
kxtran: Nit: _record(event) is called on every iteration of this loop (8 times) and we can just hoist it

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/events.py
```diff
@@ -0,0 +1,352 @@
...
+def _record(event: DCIMChangeEvent) -> Mapping[str, Any]:
+    """Return the Nautobot record attached to a changelog event."""
+    if event.record is None:
+        raise DCIMInvalidDataError(f"Nautobot event {event.object_type} has no record")
+    return event.record
+
+
+def _id(value: object, description: str) -> str:
+    """Return a required provider object identifier."""
+    if value is None or not str(value):
```
kxtran: Nit: `not str(value)` only catches the empty-string case since `None` is already handled above. Simplifying to `not value` covers all falsy inputs (None, "", 0, etc.) in one check and reads more naturally.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+    async def _get_overlay_assignment_status_id(self) -> str:
+        """Return the default status ID used for overlay assignments."""
+        status_id = await self.lookup_id_by_name(
+            "extras/statuses/", _SPECTRUM_X_DEFAULT_STATUS_NAME
+        )
+        if status_id is None:
+            raise ApplicationError(
+                f"Status {_SPECTRUM_X_DEFAULT_STATUS_NAME} not found for overlay assignment"
+            )
+        return status_id
```
kxtran: Every call to `lookup_id_by_name` in this file is immediately followed by a `None` check and an `ApplicationError` raise — 10 call sites total (7 for status, 2 for tenant, 1 for location). The existing `_get_overlay_assignment_status_id` helper already partially addresses this but only covers one of the cases. I think that replacing it with a single generic helper like:
```
async def _require_id_by_name(
    self, path: str, name: str, *, non_retryable: bool = True
) -> str:
    """Look up a Nautobot object by name, raising if not found."""
    object_id = await self.lookup_id_by_name(path, name)
    if object_id is None:
        raise ApplicationError(
            f"'{name}' not found at {path} in DCIM",
            non_retryable=non_retryable,
        )
    return object_id
```

This would replace all 10 lookup+check+raise blocks with one-liners and let `_get_overlay_assignment_status_id` be removed entirely.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+            if device_name in device_names:
+                duplicates.add(device_name)
+            else:
+                device_names.add(device_name)
+                unique_devices.append(device_data)
+
+        if duplicates:
+            raise NautobotException(f"Duplicate device names in nautobot: {duplicates}")
+
+        return unique_devices
```
kxtran: `_deduplicate_devices` builds a `unique_devices` list that's never used — if duplicates exist it raises before returning, and if none exist the list is identical to the input. Could simplify to just validate and return the original list.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+                device
+                for device in devices
+                if device.backup_enabled == device_status_filters["backup_enabled"]
+            ]
+        if device_status_filters.get("ztp_enabled") is not None:
+            devices = [
+                device
+                for device in devices
+                if device.ztp_enabled == device_status_filters["ztp_enabled"]
+            ]
```
kxtran: These four filter blocks repeat the same pattern — check if a status filter is set, then rebuild the list. Could collapse into a single pass:
```
active_filters = {
    attr: value for attr, value in device_status_filters.items() if value is not None
}
if active_filters:
    devices = [
        device for device in devices
        if all(getattr(device, attr) == value for attr, value in active_filters.items())
    ]
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+            pkey_id=str(pkey_record["id"]),
+            pkey=canonical_pkey,
+        )
+
+    async def _delete_ib_pkey_resource_if_present(self, path: str, description: str) -> None:
+        """Delete one provider resource, treating a retry-after-404 as successful."""
+        try:
+            await self.delete(path)
+        except NautobotException as error:
+            if "returned 404" not in str(error):
```
kxtran: The 404 detection here relies on matching "returned 404" in the exception message string. If the error template in `_handle_error_response` is ever reworded, this silently breaks. Consider adding a `status_code` attribute to `NautobotException` so this can check `error.status_code != 404` instead of parsing the message.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+                if (device := entry["device"])
+                and (device.get("status") or {}).get("name") in _BACKUP_SUPPORTED_STATUSES
+                and (device.get("platform") or {}).get("name") in _BACKUP_SUPPORTED_PLATFORMS
+            }
+        except (KeyError, TypeError) as exc:
+            raise DCIMInvalidDataError(
+                "Nautobot returned invalid backup-enabled device data"
+            ) from exc
+
+    async def get_os_image_versions(self, device_id: str) -> tuple[str, str, str]:
```
kxtran: Nit: `get_os_image_versions` returns a bare `tuple[str, str, str]` — callers have to know that `[0]` is intended, `[1]` is desired, and `[2]` is the ZTP IP. A `NamedTuple` would make call sites self-documenting.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+    async def get_ib_interface_records(
+        self, device_interface_pairs: list[tuple[str, str]]
+    ) -> list[IBInterfaceGuid]:
+        """Read named IB interfaces using Nautobot's REST filter semantics."""
+        records: list[IBInterfaceGuid] = []
+        for device_name, interface_name in device_interface_pairs:
+            response = await self.get(
+                "dcim/interfaces/",
+                params={"device": device_name, "name": interface_name},
+            )
```
kxtran, line 1377: `get_ib_interface_records` issues one REST call per device/interface pair. For large lists this could be slow. Consider batching with `asyncio.gather `(with a semaphore to bound concurrency), or consolidating into a single GraphQL query like `get_ib_interface_guids` already does for the same kind of data.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+        """Find all overlay PKey records in a location chain matching a PKey."""
+        matches: list[tuple[dict[str, Any], dict[str, Any], dict[str, Any]]] = []
+        for location in chain:
+            for overlay in location.get("overlays") or []:
+                for pkey_record in overlay.get("pkeys") or []:
+                    stored = str(pkey_record.get("pkey") or "")
+                    if not _IB_PKEY_VALUE_PATTERN.match(stored):
+                        continue
+                    if f"0x{int(stored, 16):04x}" == canonical_pkey:
+                        matches.append((location, overlay, pkey_record))
```
kxtran: This repeats the normalization logic from `_normalize_ib_pkey`. Extract a small `_canonicalize_pkey_value(value: str) -> str | None` that both call sites share — it prevents drift if the canonical format ever changes.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -0,0 +1,1796 @@
...
+    ) -> HostDeviceData:
+        """Update DPU inventory fields using Nautobot device and interface resources."""
+        device = await self.update_host_device(device_id=device_id, data={"serial": serial})
+        device.interfaces = []
+        for interface_id, mac_address in interface_macs.items():
+            device.interfaces.append(
+                await self.update_interface(
+                    interface_id=interface_id, data={"mac_address": mac_address}
+                )
+            )
```
kxtran: Nit: `update_dpu_device_inventory` discards the interfaces returned by `update_host_device` and replaces them with individually-patched results. Building the list separately and assigning once would make that intent clearer:
```
interfaces = [
    await self.update_interface(interface_id=iface_id, data={"mac_address": mac})
    for iface_id, mac in interface_macs.items()
]
device.interfaces = interfaces
```

components/dcim/src/nv_config_manager_dcim/registry.py
```diff
@@ -0,0 +1,116 @@
...
+)
+
+DCIM_PROVIDER_ENTRY_POINT_GROUP = "nv_config_manager.dcim"
+
+
+def _entry_points_for_group() -> tuple[EntryPoint, ...]:
+    discovered = entry_points()
+    if hasattr(discovered, "select"):
+        return tuple(discovered.select(group=DCIM_PROVIDER_ENTRY_POINT_GROUP))
+    return tuple(discovered.get(DCIM_PROVIDER_ENTRY_POINT_GROUP, ()))
```
kxtran: The `hasattr(discovered, "select") / .get()` fallback was written for Python <3.12, but this package requires >=3.13. The old fallback was the sole mypy error. Replaced with a single `entry_points(group=...)` call.

components/dcim/src/nv_config_manager_dcim/registry.py
```diff
@@ -0,0 +1,116 @@
...
+        providers[entry_point.name] = provider
+    return providers
+
+
+def get_dcim_provider(name: str) -> DCIMProvider:
+    """Return an installed provider by explicit entry-point name."""
+    try:
+        return discover_dcim_providers()[name]
+    except KeyError as exc:
+        installed = ", ".join(sorted(discover_dcim_providers())) or "none"
```
kxtran: `get_dcim_provider` is calling twice here and we can hoist it

components/dcim/src/nv_config_manager_dcim/models.py
```diff
@@ -0,0 +1,377 @@
...
+        required = ("provider", "operation", "object_type", "object_id", "timestamp")
+        missing = [field for field in required if not str(data.get(field, "")).strip()]
+        if missing:
+            raise DCIMInvalidDataError(
+                "DCIM event is missing required field(s): " + ", ".join(missing)
+            )
+        if str(data.get("contract_version", "")).strip() != DCIM_EVENT_CONTRACT_VERSION:
+            raise DCIMInvalidDataError("Unsupported DCIM event contract version")
+        operation = str(data["operation"])
+        if operation not in {"create", "update", "delete"}:
```
kxtran: Nit: the `{"create", "update", "delete"}` set literal is rebuilt on every `from_dict` call. Hoist it to a module-level `frozenset` so it's allocated once.

components/dcim/src/nv_config_manager_dcim/workflow_models.py
```diff
@@ -0,0 +1,151 @@
...
+    @computed_field  # type: ignore[misc]
+    @property
+    def intended_config_file(self) -> str:
+        if self.platform in {Platform.ARISTA_EOS, Platform.MLNX_OS}:
+            return "full-config"
+        if self.platform in {Platform.CUMULUS_LINUX, Platform.NV_OS}:
+            return "startup.yaml"
+        if self.platform == Platform.UFM:
+            return ""
+        raise NotImplementedError(f"No configuration path for platform {self.platform}")
```
kxtran: Replace the `if/elif/raise NotImplementedError` chain with `match/case` and `assert_never` in the default branch. As-is, adding a new Platform member without updating this property compiles cleanly and only fails at runtime. `assert_never` makes `mypy` catch the missing case at type-check time.

```
        match self.platform:
            case Platform.ARISTA_EOS | Platform.MLNX_OS:
                return "full-config"
            case Platform.CUMULUS_LINUX | Platform.NV_OS:
                return "startup.yaml"
            case Platform.UFM:
                return ""
            case _ as unreachable:
                assert_never(unreachable)
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow_models.py
```diff
@@ -0,0 +1,119 @@
...
+    """Map one Nautobot display name to the SDK's stable slug."""
+    return name.lower().replace(" ", "-")
+
+
+def _extract_site(device: dict[str, Any]) -> str:
+    """Find the Site record in a Nautobot location hierarchy."""
+    location = device.get("location")
+    while location:
+        if (location.get("location_type") or {}).get("name") == "Site":
+            return str(location["name"])
```
kxtran: If a Nautobot Site location has name: `null` in the GraphQL response, `str(location["name"])` silently produces the string "None" and returns it as a valid site name. Downstream consumers would then operate against a bogus site without any error.

`_extract_site` uses `str(location["name"])`, and `network_device_from_nautobot_graphql / host_device_from_nautobot_graphql` use `str(role["name"])`, `str(platform["name"])`, `str(device_type["model"])`. If any of these values are null in the GraphQL response, `str(None)` silently produces the string "None" and passes it downstream as a valid value. For platform, this then hits `Platform("none")` which raises a raw `ValueError` from `StrEnum` instead of a domain-specific `DCIMInvalidDataError`.

Add a shared helper to validate required string fields:
```
def _require_str(data: dict[str, Any], key: str, label: str) -> str:
    value = data.get(key)
    if not isinstance(value, str) or not value:
        raise DCIMInvalidDataError(f"{label} is missing required field '{key}'")
    return value
```
Then replace every `str(d["field"])` call with `_require_str(d, "field", "label")` — in `_extract_site`, `network_device_from_nautobot_graphql`, and `host_device_from_nautobot_graphql`.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow_models.py
```diff
@@ -0,0 +1,119 @@
...
+    device = interface.get("device") or (interface.get("module") or {}).get("device")
+    if not device:
+        raise DCIMInvalidDataError(f"Interface {interface.get('id')} has no device associated")
+    mac_address = interface.get("mac_address")
+    vrf = interface.get("vrf") or {}
+    return InterfaceData(
+        id=str(interface["id"]),
+        name=str(interface["name"]),
+        host=str(device.get("name") or device.get("id")),
+        mac_address=str(netaddr.EUI(mac_address)) if mac_address else None,
```
kxtran: If Nautobot returns a non-empty but malformed MAC address, `netaddr.EUI(mac_address)` raises `netaddr.core.AddrFormatError`. Callers catching `DCIMError` won't see it — the third-party exception propagates unhandled.

Wrap the call so it stays within the SDK's error contract:

```
try:
    parsed_mac = str(netaddr.EUI(mac_address)) if mac_address else None
except netaddr.core.AddrFormatError as exc:
    raise DCIMInvalidDataError(
        f"Interface {interface.get('name')} has invalid MAC address: {mac_address}"
    ) from exc
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow_models.py
```diff
@@ -0,0 +1,119 @@
...
+    HostDeviceData,
+    InterfaceData,
+    NetworkDeviceData,
+    Platform,
+)
+
+
+def _slugify(name: str) -> str:
+    """Map one Nautobot display name to the SDK's stable slug."""
+    return name.lower().replace(" ", "-")
```
kxtran: Nit: if a Nautobot display name ever has leading or trailing spaces, the current implementation would produce a slug with leading/trailing hyphens (e.g. " Cumulus Linux " → "-cumulus-linux-"). A .strip() before the replace would prevent that:
```
return name.strip().lower().replace(" ", "-")
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/queries.py
```diff
@@ -0,0 +1,167 @@
...
+        if isinstance(selection, FragmentSpreadNode):
+            names.add(selection.name.value)
+        nested_selection_set = getattr(selection, "selection_set", None)
+        if nested_selection_set is not None:
+            names.update(_fragment_spreads(nested_selection_set))
+    return names
+
+
+@cache
+def _load_graphql_source(filename: str) -> str:
```
kxtran, line 79: the func can hit `RecursionError `on circular imports

Mutually-importing `.graphql` files recurse infinitely — @cache can't help because the first call hasn't returned yet. A transient sentinel set gives a clear error instead:

```
_LOADING: set[str] = set()
@cache
def _load_graphql_source(filename: str) -> str:
    if filename in _LOADING:
        raise ValueError(f"Circular GraphQL import detected: {filename!r}")
    _LOADING.add(filename)
    try:
        ...  # existing body unchanged
    finally:
        _LOADING.discard(filename)
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+                }
+            )
+            config_context = device.get("config_context") or {}
+            firmware_version = config_context.get("intended-firmware", {}).get("version")
+            intended_config = managed_device.get("intended_config")
+            config_store_instance = None
+            if intended_config:
+                config_store_instance = re.sub(
+                    "ui", "api-mtls", intended_config["config_store_instance"]
+                )
```
kxtran: `re.sub("ui", "api-mtls", ...)` is unanchored -- it will replace any occurrence of "ui" in the string, including inside words like "circuit" or "requisition". Should this target a specific URL component instead?
```
# e.g. if swapping a subdomain:
config_store_instance = intended_config["config_store_instance"].replace("://ui.", "://api-mtls.", 1)
# or if it's a standalone path/host segment:
config_store_instance = re.sub(r"\bui\b", "api-mtls", intended_config["config_store_instance"])
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+            return False
+
+    @staticmethod
+    def is_valid_location_id(value: str) -> bool:
+        """Return whether a location identifier is a canonical Nautobot UUID."""
+        try:
+            return str(UUID(value)) == value
+        except (ValueError, TypeError, AttributeError):
+            return False
+
```
kxtran, line 249: `is_valid_device_id` and `is_valid_location_id` are identical implementations. Extract a shared `_is_canonical_uuid` helper

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+        location = (result.get("data") or {}).get("location")
+        if location is None:
+            return None
+        selections = self._parameter_selections([location], "location")
+        return selections[0]
+
+    async def get_device_metadata(self, device_id: str) -> DeviceMetadata | None:
+        """Return normalized metadata for a Nautobot device UUID."""
+        result = await self.graphql_query(
+            load_graphql_query("provider/devices.graphql", "GetDeviceMetadata"),
```
kxtran: should we move this inline `load_graphql_query()` to the module level constants for consistency?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+            not isinstance(device, dict)
+            or not isinstance(device.get("id"), str)
+            or not isinstance(device.get("name"), str)
+        ):
+            raise DCIMInvalidDataError("Nautobot returned invalid device choice data")
+        return DCIMDeviceSelection(id=device["id"], name=device["name"])
+
+    async def get_managed_device_metadata(self, page_size: int = 100) -> list[DeviceMetadata]:
+        """Return normalized metadata for every NVCM-managed Nautobot device."""
+        query = load_graphql_query("provider/devices.graphql", "ListManagedDeviceMetadata")
```
kxtran: should we move this inline `load_graphql_query()` to the module level constants for consistency?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+            )
+        except (KeyError, TypeError) as exc:
+            raise DCIMInvalidDataError(
+                f"Nautobot returned incomplete ZTP device data for {device_id}"
+            ) from exc
+
+    async def get_device_serial(self, device_id: str) -> str:
+        """Return a Nautobot device serial number for ZTP validation."""
+        result = await self.graphql_query(
+            load_graphql_query("provider/devices.graphql", "GetDeviceSerial"),
```
kxtran: should we move this inline load_graphql_query() to the module level constants for consistency?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+
+    async def mark_ztp_device_provisioned(self, device_id: str) -> None:
+        """Mark a Nautobot device provisioned after successful ZTP."""
+        await self.set_status_provisioned(device_id)
+
+    async def get_render_data(self, request: RenderDataRequest) -> RenderData:
+        """Return the Nautobot data set consumed by the template engine."""
+        device_id = request.device_id
+        device_data = await self.graphql_query(
+            load_graphql_query("query_config_data_by_device_id_v2.graphql"),
```
kxtran: should we move this inline load_graphql_query() to the module level constants for consistency?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+            while location["location_type"]["name"] != "Site":
+                location = location["parent"]
+            location_name = location["name"]
+        except (KeyError, TypeError) as exc:
+            raise DCIMInvalidDataError(
+                f"Nautobot returned incomplete render location data for {device_id}"
+            ) from exc
+
+        location_data = await self.graphql_query(
+            load_graphql_query("query_location_data.graphql"), {"location": location_name}
```
kxtran: should we move this inline load_graphql_query() to the module level constants for consistency?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+        location = location.get("parent")
+
+    return DeviceMetadata(
+        device_id=data["id"],
+        name=data["name"],
+        site=site or "Unknown",
+        platform=data.get("platform", {}).get("name") if data.get("platform") else None,
+        role=data.get("role", {}).get("name") if data.get("role") else None,
+        rack=data.get("rack", {}).get("name") if data.get("rack") else None,
+        primary_ip4=data.get("primary_ip4", {}).get("host") if data.get("primary_ip4") else None,
```
kxtran: nit: each optional field calls `.get()` twice. Walrus operator cleans this up:
```
platform=(p.get("name") if (p := data.get("platform")) else None),
```

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+            connected_interface = interface.get("connected_interface")
+            if not isinstance(connected_interface, dict):
+                normalized.append(
+                    IntendedInterfaceNeighbor(
+                        name=interface["name"],
+                        tags=tuple(
+                            str(tag["name"])
+                            for tag in interface.get("tags", [])
+                            if isinstance(tag, dict) and tag.get("name") is not None
+                        ),
```
kxtran: The tag-extraction comprehension is duplicated with [L295-299](https://github.com/NVIDIA/nv-config-manager/pull/226/changes#diff-88f96a3fb17d4785a80823426d55673e94835b553226765c5bb93c04c3f5191eR295-R299). Extract a small helper to DRY it up.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+                return await self.get_module_bay_parent_device_id(str(parent_module_bay["id"]))
+
+        url = termination.get("url")
+        if not isinstance(url, str) or not url:
+            raise DCIMInvalidDataError("Nautobot cable termination has no device or REST URL")
+        endpoint = url.split("/api/", maxsplit=1)[-1].lstrip("/")
+        resolved = await self.get(endpoint)
+        if not isinstance(resolved, Mapping):
+            raise DCIMInvalidDataError("Nautobot returned an invalid cable termination")
+        return await self.get_cable_termination_device_id(resolved)
```
kxtran: `get_cable_termination_device_id` calls itself recursively with no depth guard. If the resolved termination also lacks a device/module and has a URL, this loops forever. Add a max-depth counter or convert to an iterative loop?

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/provider.py
```diff
@@ -0,0 +1,823 @@
...
+    """Normalize a provider-owned TLS verification setting."""
+    if isinstance(value, bool):
+        return value
+    if value is None or str(value).strip() == "":
+        return True
+    normalized = str(value).strip()
+    if normalized.lower() in {"true", "yes", "1"}:
+        return True
+    if normalized.lower() in {"false", "no", "0"}:
+        return False
```
kxtran: Nit: the `strip()` and `lower()` are computed twice. Let's hoist it
```
if value is None:
    return True
normalized = str(value).strip()
if not normalized:
    return True
lowered = normalized.lower()
```

src/nv_config_manager/config_store/api/device_ids.py
```diff
@@ -0,0 +1,48 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Provider-aware Config Store device identifier validation."""
+
+from typing import TYPE_CHECKING
+
+from fastapi import HTTPException, Request
+
+if TYPE_CHECKING:
+    from nv_config_manager.config_store.core.device_cache_redis import DeviceCacheService
```
kxtran, line 24: nit: can we use ` __future__` annotations for consistency instead?

src/nv_config_manager/dcim/registry.py
```diff
@@ -0,0 +1,224 @@
...
+    """Return installed providers from the standalone SDK discovery layer."""
+    return discover_sdk_dcim_providers()
+
+
+def configured_dcim_provider_name(config: ConfigParser | None = None) -> str:
+    """Return the service-selected provider, defaulting to Nautobot 2.x."""
+    if config is None:
+        from nv_config_manager.common.config import load_config  # avoid circular import
+
+        config = load_config()
```
kxtran: The `if config is None: from ... import load_config; config = load_config()` block is repeated multiple times. Extract it into a helper.

src/nv_config_manager/dcim/registry.py
```diff
@@ -0,0 +1,224 @@
...
+        f"dcim.{provider_name}",
+    ):
+        if config.has_section(section):
+            for key, value in config.items(section):
+                if key != "provider":
+                    settings[key] = value
+    return settings
+
+
+def get_dcim_provider(config: ConfigParser | None = None) -> Any:
```
kxtran, line 109: nit: 
```suggestion
def get_dcim_provider(config: ConfigParser | None = None) -> DCIMProvider:
```

src/nv_config_manager/dcim/registry.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def create_dcim_workflow_client(config: ConfigParser | None = None) -> DCIMClient:
+    """Return a service-managed client compatible with legacy workflow callers."""
+    return cast(DCIMClient, _ManagedDCIMClient(create_dcim_client(config)))
+
+
+def create_dcim_parameter_client(config: ConfigParser | None = None) -> DCIMClient:
+    """Return a service-managed client compatible with legacy API callers."""
+    return cast(DCIMClient, _ManagedDCIMClient(create_dcim_client(config)))
```
kxtran: these functions are identical. Are they going to be diverged in the future?

src/nv_config_manager/render/dispatch.py
```diff
@@ -106,59 +114,89 @@
...
+                event.actor,
+                event.timestamp,
+                dcim_client=dcim_client,
+            )
+            for failure in failures:
+                self.logger.info(
+                    "Render request for %s was not queued: %s",
+                    failure["device_uuid"],
+                    failure["error"],
+                )
```
ryanheffernan, line 171: should we throw an exception or otherwise return a failure condition here?

src/nv_config_manager/render/render.py
```diff
@@ -55,141 +54,116 @@ async def _update_nautobot_plugin(  # pylint: disable=too-many-arguments
...
-                    user,
-                    commit_message,
+            if not files:
+                logger.info("No entrypoint templates matched for %s", device_uuid)
+                return []
+            files = {key.split("/")[-1].replace(".j2", ""): value for key, value in files.items()}
+        except Exception as exc:
+            raise RenderException(f"Failed to render entrypoints for {device_uuid}: {exc}") from exc
+
+        csclient = config_store_client()
```
kxtran: The try/except on L92-109 only covers the render phase — everything after (config-store persist, DCIM sync, resync loop) runs unprotected. A failure there surfaces as a raw exception with no device context. Wrap L111-169 in its own handler.

src/nv_config_manager/dcim/registry.py
```diff
@@ -0,0 +1,224 @@
...
+    ``[dcim]`` and ``[dcim.options]`` are the portable service configuration
+    surface. A provider-specific ``[dcim.<provider>]`` section takes
+    precedence without leaking ``ConfigParser`` into the provider SDK.
+    """
+    provider_name = provider_name or configured_dcim_provider_name(config)
+    settings: dict[str, str] = {}
+    for section in (
+        "dcim",
+        "dcim.options",
+        f"dcim.{provider_name}",
```
ryanheffernan, line 100: should we also reat the [nautobot] section here so we copy the legacy nautobot config values as well?

installer/src/nv_config_manager_installer/schema.py
```diff
@@ -932,14 +1017,30 @@ class NVConfigManagerInstallConfig(BaseModel):
...
+                raise ValueError(
+                    "dcim.server is required when dcim.provider is not the built-in nautobot-2x"
+                )
+            if (
+                self.secrets.method == SecretsMethod.ESO
+                and not self.secrets.vault.paths.dcim.enabled
+            ):
+                raise ValueError(
+                    "secrets.vault.paths.dcim.enabled must be true for an external DCIM provider with ESO"
+                )
```
ryanheffernan, line 1074: should we also validate `secrets.vault.paths.nats.enabled` here?

src/nv_config_manager/temporal/ngc/activities/device.py
```diff
@@ -16,90 +16,67 @@
...
+                device_role=(
+                    interface.connected_device.role.lower().replace(" ", "-")
+                    if interface.connected_device.role
+                    else None
+                ),
+                device_rack=interface.connected_device.rack,
+                device_position=interface.connected_device.position,
+            )
+            for interface in interfaces
+            if interface.connected_device
```
kxtran: Nit: the `interface.connected_device` access is repeated multiple times after the if guard already confirms it's truthy. A walrus (`if (dev := interface.connected_device)`) would bind it once and cut the noise.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
zsblevins: Nautobot enforces that there will be no circular chain, also note that our GraphQL queries don't infinitely recurse, they stop 3 levels deep for the parent pull. Probably harmless to add a depth check, but should be unnecessary

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/dhcp.py
zsblevins: legacy, meant to remove it and forgot to get back to it, good catch :)

src/nv_config_manager/render/dispatch.py
zsblevins: Agreed that partial enqueue failures should be surfaced more clearly. We’re deferring this to a dedicated enqueue cleanup PR because that path has become exceptionally slow and needs broader work around batching, failure propagation, and retry behavior. We’re keeping the existing logging behavior in this DCIM-neutrality PR to avoid changing enqueue semantics here.

installer/src/nv_config_manager_installer/schema.py
zsblevins: We intentionally don’t validate the NATS Vault path here. NATS is no longer coupled to the selected DCIM provider; keeping those concerns separate makes it possible for providers such as NetBox to supply their own installer integration without inheriting Nautobot’s NATS requirements. Any NATS secret validation should be based on whether NATS is configured for the deployment, independently of `dcim.provider`.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
zsblevins: This is a valid performance concern, but we’re deferring it to a targeted InfiniBand cleanup PR. We currently lack integration coverage for this path and don’t have a test environment populated with representative InfiniBand data, so changing its concurrency or query behavior in this PR would be difficult to validate safely. We’ll treat batching or GraphQL consolidation as a future enhancement once that coverage is available.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
zsblevins: We’re intentionally retaining `TypeError` here because malformed provider data can raise it during normalization before Pydantic validation—for example, when a field expected to be integer-like is instead a list or mapping. This boundary should translate those malformed payloads into `DCIMInvalidDataError`, and the chained exception preserves the original cause for debugging. Narrowing this would require validating every individual conversion without changing the resulting contract.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
zsblevins: We’re not adding a depth guard here. Nautobot enforces an acyclic location hierarchy, and the GraphQL response includes only a finite parent depth, so this function is walking an already-bounded response rather than recursively querying Nautobot. A true cycle would require a synthetic self-referential Python mapping that cannot be produced by the JSON GraphQL response.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/client.py
zsblevins: `load_graphql_query()` returns `GraphQLOperation`, which is a `str` subclass carrying the selected `operation_name`. The transport remains typed as `str` so existing callers can still submit ordinary query strings; `getattr()` distinguishes the enriched loader result without breaking those callers. This is required when a source document contains multiple operations, and the transport behavior is covered by a focused test.

kxtran: gotcha, the `query` is defined as `str` type threw me off.

components/dcim-nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/render.py
kxtran: I see, thanks for the context.

components/dcim/src/nv_config_manager_dcim/workflow_models.py
```diff
@@ -0,0 +1,162 @@
...
+
+from enum import StrEnum
+from typing import Any, ClassVar, NamedTuple, assert_never
+
+from pydantic import BaseModel, computed_field
+
+from nv_config_manager_dcim.errors import DCIMInvalidDataError
+
+
+class Platform(StrEnum):
```
ryanheffernan, line 27: will need to update the references from dcim_name here: https://github.com/dsx-ai-factory/nv-config-manager/blob/1e4c0393004716cb8facf9630364db63d02586db/src/nv_config_manager/temporal/ngc/workflows/cable_validation.py#L242

ui/src/app/workflows/columns.tsx
```diff
@@ -215,7 +215,10 @@ function DeviceNameCell({ workflow }: { workflow: WorkflowColumns }) {
       param="device_name"
       value={deviceNameValue}
     >
-      {renderDeviceNameField(workflow, config?.nautobotUrl)}
+      {renderDeviceNameField(
+        workflow,
+        config?.dcimProvider === "nautobot" ? config.dcimUrl : undefined
+      )}
```
ryanheffernan, line 226: the default helm provider value will clash with this

https://github.com/dsx-ai-factory/nv-config-manager/blob/46e48bb86a6b071bef67e68de2098bc3afa620cc/deploy/helm/values.yaml#L44

src/nv_config_manager/temporal/api/parameter_v1.py
```diff
@@ -557,34 +318,17 @@ async def get_diagnostics_commands(
...
-
-    device_data = data["data"]["device"]
-
-    config_context = device_data.get("config_context", {}) or {}
-
-    secret_versions = config_context.get("secrets_versions", {})
+        secret_versions = await client.get_device_secret_versions(device_id)
 
     secrets = []
     if isinstance(secret_versions, dict):
```
ryanheffernan: the secret methods are declared with Mapping[str, str] but this accepts only dict

https://github.com/dsx-ai-factory/nv-config-manager/blob/46e48bb86a6b071bef67e68de2098bc3afa620cc/components/dcim/src/nv_config_manager_dcim/api.py#L177-L179

components/network-templates/src/nv_config_manager_templates/filters/device.py
```diff
@@ -325,220 +339,148 @@ def console_server_ports(
...
+    for instance in routing_instances:
+        if vrf not in instance.vrfs and not (vrf == "default" and len(routing_instances) == 1):
+            continue
+        peers = tuple(peer for peer in instance.peers if peer.source_vrf == vrf)
+        interface = _required(
+            instance.router_id_interface,
+            "device.routing.bgp_instances[].router_id_interface",
+            value,
+        )
+        return BGPLocalConfig(
```
susanhooks: I think this could continue on vrf "default" , and then peers would be an empty tuple, not matching a further down the list match.

kxtran (approved):
🚀

PR #248, opened 2026-08-20
fix(dcim): restore the Nautobot request budget and make it configurable

dmathur0:
## Summary

Rebased onto `main` after the pluggable DCIM split. The original branch predated the split and targeted six files that no longer exist, so the change is reapplied against `plugins/dcim/nautobot-2x/`.

**The regression (first commit).** `NautobotWorkflowClient.graphql_query` defaults `timeout` to 10s. `NautobotDCIMClient` inherits from it, so after the split *every* provider GraphQL call — including the DHCP and render paths that previously had a wider budget — was capped at 10s regardless of the client's own 30s default. That is what test01 is timing out on. It now defaults to `None` and inherits the client budget; explicit per-request timeouts still win.

**The configurable budget (second commit).** Adds an optional `[dcim] timeout` that the `nautobot-2x` provider normalizes and passes to its client. When unset, the client keeps its built-in 30s default (`DEFAULT_TIMEOUT`). A non-numeric or non-positive value raises `DCIMProviderConfigurationError` at client construction rather than silently reverting to the default. Exposed as `dcim.timeout` in the chart and rendered into all three `nv-config-manager.ini` paths (Kubernetes secrets, ESO, Vault Agent).

Because settings now flow through `provider_settings`, `[dcim]`, `[dcim.options]`, `[dcim.nautobot-2x]`, and legacy `[nautobot]` all work, with the usual precedence.

### Dropped from the original PR

- The `pynautobot` / `TimeoutHTTPAdapter` render commit — `pynautobot_client` no longer exists; render goes through the async client.
- The per-subclass DHCP 60s default — DHCP is now a mixin on the single `NautobotDCIMClient`, so there is one client and one default.
- The `[nautobot]` Helm plumbing, in favor of the canonical `[dcim]` surface. Legacy `[nautobot] timeout` is still honored in code for 1.x deployments.

## Test plan

- [x] `uv run pytest src/tests/dcim` — 68 passed, including new coverage for the timeout reaching the client, defaulting when unset/blank, legacy `[nautobot]` support, rejection of unusable values, the MCP adapter sharing the budget, and a regression guard that GraphQL calls use the client budget by default
- [x] `uv run pytest src/tests/dcim src/tests/temporal src/tests/ztp src/tests/dhcp` — 1317 passed (one pre-existing Docker-dependent Kea failure, unrelated)
- [x] `uv run ruff check` / `ruff format --check` clean; `mypy` error count unchanged from `main` (13 pre-existing)
- [x] `helm template` for `secrets.method` in {`eso`, `kubernetes`, `vault-agent`}: no `timeout` under `[dcim]` when unset, `timeout = 120` when `dcim.timeout=120`

## Notes

Zoe — this only restores and parameterizes the budget; the pagination change is still outstanding and not included here.

## Summary by CodeRabbit

- **New Features**
  - Added optional configurable Nautobot request timeouts through Helm deployment settings.
  - Configured timeouts now apply consistently to HTTP, HTTPS, and GraphQL requests.
  - Requests without an explicit timeout use the configured client timeout or provider default.

- **Bug Fixes**
  - Added fallback handling for missing or blank timeout settings.
  - Invalid, non-positive, or nonnumeric timeout values are rejected with configuration errors.

## Kind Integration

Kind integration completed with **success** for [`72779c6`](https://github.com/dsx-ai-factory/nv-config-manager/commit/72779c656dfb8ad9ff1fe1541480c8dd7c8f45c3) in [this workflow run](https://github.com/dsx-ai-factory/nv-config-manager/actions/runs/34919400756).


deploy/helm/values.yaml
```diff
@@ -493,6 +493,11 @@ externalServices:
     version: 2.3.5
     maxWorkers: 8
     retries: 3
+    # Client HTTP/GraphQL timeout in seconds for services that talk to Nautobot
+    # (dhcp-refresh, render, config-store, ztp, temporal). DHCP GraphQL on large
+    # aggregate cells regularly takes ~60s; 120 leaves headroom without waiting
+    # on a hung Nautobot. Override per cell if needed (e.g. 180 on agg-prod).
+    timeout: 120
```
ryanheffernan: this will get overridden by the render client code: https://github.com/NVIDIA/nv-config-manager/blob/efad1e1b9a0a633b82771ebdc0327c3c9ec25db8/src/nv_config_manager/common/config.py#L678

daveshm: ty, fixing that. although that query may be faster

dmathur0: Good catch — `pynautobot_client()` was still mounting `TimeoutHTTPAdapter(timeout=10)`, so render never saw `[nautobot] timeout`.

It now uses `NautobotClient.timeout_from_config()`, same as the async clients. Unset INI still falls back to 30s (the shared default), and Helm 120s applies to render as well.

deploy/helm/templates/kubernetes-secrets.yaml
```diff
@@ -126,6 +126,9 @@ spec:
               public_url = {{ include "nv-config-manager.dcimPublicUrl" . }}
               token = $(cat /secrets/dcim-token/{{ $dcimTokenSecretKey }})
               verify = {{ include "nv-config-manager.dcimVerify" . }}
+              {{- if .Values.dcim.timeout }}
+              timeout = {{ .Values.dcim.timeout }}
+              {{- end }}
```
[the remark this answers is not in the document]
dmathur0: Not applying this one — `0` is the chart's intended sentinel for "unset", not an invalid value to be surfaced.

`deploy/helm/values.yaml` documents and ships `dcim.timeout: 0` as the default, meaning "keep the provider's built-in 30s budget". Because `_parse_timeout` in `provider.py` raises on any value `<= 0`, rendering a configured `0` into the INI would emit `timeout = 0` for the default values file, and every out-of-the-box install would fail provider configuration at startup.

The sentinel also matches the established convention for the sibling keys: `dcimCacheRefreshInterval` and `dcimCacheTtl` in `_helpers.tpl` use `| default`, which treats `0` as unset the same way.

Genuinely invalid input still fails loudly — a negative such as `-5` is truthy in Go templates, so it renders and the provider rejects it with a clear error.

One thing from this comment is worth keeping, though: there are three of these conditions, not two. `deploy/helm/templates/_vault-agent.tpl#L311-L313` has the same block and was not listed. Any future change here has to cover all three to keep the consul-template body in sync with `vault-secrets.yaml`.

plugins/dcim/nautobot-2x/src/nv_config_manager_dcim_nautobot_2x/workflow.py
```diff
@@ -163,13 +163,21 @@ def __init__(self, connection_config: Mapping[str, Any]) -> None:
             nautobot_url=str(connection_config["server"]),
             token=str(connection_config["token"]),
             verify=connection_config["verify"],
+            timeout=connection_config.get("timeout"),
             headers=connection_config.get("headers"),
         )
 
     async def graphql_query(
-        self, query: str, variables: dict[str, Any] | None = None, timeout: int | None = 10
+        self, query: str, variables: dict[str, Any] | None = None, timeout: int | None = None
```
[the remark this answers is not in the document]
dmathur0: Not applying this one — there is no 60-second DHCP timeout requirement in this repository.

`git log -S 'timeout=60' -- plugins/dcim/nautobot-2x/` returns nothing, so such a value was never present and was not removed by this PR. `dhcp.py` is not part of this PR's diff at all.

Omitting `timeout` at those call sites is the intended behavior and the point of this change: the base client resolves `self._timeout` from the deployment's configured budget, falling back to `DEFAULT_TIMEOUT` only when nothing is configured. Hardcoding `60` would make the DHCP path ignore the configurable budget this PR adds.

zsblevins (approved):

PR #256, opened 2026-08-24
feat(workflows): add pluggable workflow registration

kxtran:
## Description

Introduces the `nv-config-manager-workflows` package: a Python setuptools entry-point mechanism (nv_config_manager.workflows) for discovering and registering Temporal workflow plugins independently of the core codebase.

- Discovery (discovery.py) — scans installed distributions for the `nv_config_manager.workflows` entry-point group, resolving each to a WorkflowPluginDescriptor and failing loudly on duplicate names, load errors, or name/descriptor mismatches.
- Contract (contract.py) — defines the attribute contract the registry reads from workflow/activity classes (name, description, input class, API endpoint, MCP exposure, required activities) via duck-typed accessors rather than isinstance checks, so any class satisfying the contract works. Temporal identity (workflow type / activity name) is read from the SDK's own @workflow.defn/@activity.defn definitions rather than duck-typed, so an SDK rename fails fast at import instead of silently.
- Validation (validation.py) — checks metadata completeness and cross-plugin uniqueness (names, API endpoints, CLI names) before workflows/activities are validated, catching conflicts between plugins before registration.
- Registry (registry.py) — the runtime registry combining discovery + validation to expose the built-in plugin and any installed third-party plugins to the worker/API/CLI/MCP consumers.

Includes unit tests for descriptor, registry, discovery, contract metadata, and validation (packages/workflows/tests/registration/).

## Validation

- [ ] Standard CI passes.

```text
/ok to test <sha>
```

```text
/kind test
```

only while debugging.

Kind integration completed with **success** for [`4ef4f9f`](https://github.com/dsx-ai-factory/nv-config-manager/commit/4ef4f9f8a3d92ae7e30dd820eeff230eb8d33092) in [this workflow run](https://github.com/dsx-ai-factory/nv-config-manager/actions/runs/32773509690).

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a reusable workflow framework with stages, approvals, retries, dependencies, outputs, and failure tracking.
  * Enabled API access for device, deployment, validation, backup, upgrade, and provisioning workflows.
  * Added workflow metadata, resource locking, search attributes, Markdown summaries, and compressed stage history.
  * Added discovery, validation, conflict detection, and diagnostics for installed workflow extensions.

* **Bug Fixes**
  * Improved log safety by escaping structured and formatted values.

* **Refactor**
  * Preserved existing imports through compatibility modules while centralizing shared workflow functionality.


kxtran:
/ok to test [4ef4f9f](https://github.com/dsx-ai-factory/nv-config-manager/pull/256/commits/4ef4f9f8a3d92ae7e30dd820eeff230eb8d33092)

polarweasel (dismissed):

packages/workflows/src/nv_config_manager_workflows/registration/registry.py
```diff
@@ -0,0 +1,92 @@
...
+        ordered = dict(sorted(discovered.items()))
+        validate_plugins(ordered)
+
+        all_workflows = list(dict.fromkeys(w for d in ordered.values() for w in d.workflows))
+        all_activities = list(dict.fromkeys(a for d in ordered.values() for a in d.activities))
+
+        return cls(
+            all_workflows=all_workflows,
+            all_activities=all_activities,
+            api_workflows=[w for w in all_workflows if workflow_has_complete_metadata(w)],
```
zsblevins: I don't love `workflow_has_complete_metadata` here, I think `workflow_api_enabled` would be a better mechanism. Several child workflows are intentionally not API enabled, that doesn't really mean they're incomplete, it means they are not meant for direct invocation (for security reasons for the most part)

packages/workflows/tests/registration/test_contract_metadata.py
```diff
@@ -0,0 +1,326 @@
...
+    def test_undeclared_endpoint_reads_as_none(self) -> None:
+        assert workflow_api_endpoint(BareWorkflow) is None
+        assert normalized_api_endpoint(BareWorkflow) is None
+
+    def test_a_trailing_slash_does_not_make_a_second_endpoint(self) -> None:
+        assert normalized_api_endpoint(TrailingSlashWorkflow) == normalized_api_endpoint(
+            FullyDeclaredWorkflow
+        )
+
+    def test_root_endpoint_survives_normalization(self) -> None:
```
zsblevins: I can't think of valid use cases for a root endpoint to be the invocation path for a workflow, I'd recommend making this an error instead

packages/workflows/tests/registration/test_contract_metadata.py
```diff
@@ -0,0 +1,326 @@
...
+        with pytest.raises(WorkflowRegistrationError, match="the plugin's own diagnosis"):
+            workflow_cli_name(DiagnosingCliNameWorkflow)
+
+
+class TestMcpExposure:
+    def test_mcp_is_opt_in(self) -> None:
+        assert not workflow_mcp_enabled(BareWorkflow)
+        assert workflow_mcp_enabled(FullyDeclaredWorkflow)
+
+    def test_a_non_bool_flag_still_reads_as_a_bool(self) -> None:
```
zsblevins: Commenting here, but this is a broader comment. The WorkflowMetadata should have strict typing as this feeds the Swagger definitions which must not change as part of this effort. We can add fields but we cannot change the type of existing fields nor remove any existing fields, otherwise we'd need to bump our Major version to 2.x

We don't really need flexibility here, workflows should be inheriting the base classes and mixins we already define, we don't need to allow plugins to bypass those. They may opt to add their own additional mixins, but it's fine to remain strict about the core workflow metadata and definitions in general.

ryanheffernan:
@coderabbitai review

packages/workflows/src/nv_config_manager_workflows/registration/registry.py
```diff
@@ -0,0 +1,92 @@
...
+        Raises:
+            WorkflowRegistrationError: Discovery or validation rejected the
+                installed set; see the subclasses in ``errors`` for which.
+        """
+        discovered = discover_workflow_plugins() if plugins is None else dict(plugins)
+        ordered = dict(sorted(discovered.items()))
+        validate_plugins(ordered)
+
+        all_workflows = list(dict.fromkeys(w for d in ordered.values() for w in d.workflows))
+        all_activities = list(dict.fromkeys(a for d in ordered.values() for a in d.activities))
```
ryanheffernan, line 85: should we also collect d.schedulers here and expose them through the registry?

pyproject.toml
```diff
@@ -50,7 +50,7 @@ dependencies = [
     "nkeys>=0.2.0",
     # Temporal
     # Security: Fix rustls-webpki vulnerability in Temporal bridge (GHSA-82j2-j2ch-gfr8)
-    "temporalio[opentelemetry]>=1.29.0",
+    "temporalio[opentelemetry]>=1.29.0,<1.32.0",
```
ryanheffernan, line 53: should we also add a public ci run and lint step for the new package?

packages/workflows/src/nv_config_manager_workflows/registration/validation.py
```diff
@@ -0,0 +1,290 @@
...
+        )
+
+
+def _require_endpoint_wellformed(workflow: type, label: str) -> None:
+    """Reject an API path the router could not serve."""
+    endpoint = workflow_api_endpoint(workflow)
+    if endpoint is None:
+        return
+    _require_text(endpoint, "workflow_api_endpoint", label)
+    if not endpoint.startswith("/") or any(character.isspace() for character in endpoint):
```
ryanheffernan, line 255: the endpoint '///' would pass this validation, maybe worth some additional validation to make sure there is at least one path component?

packages/workflows/tests/registration/test_contract_metadata.py
kxtran: done ✅

packages/workflows/src/nv_config_manager_workflows/registration/registry.py
kxtran: ✅

packages/workflows/src/nv_config_manager_workflows/registration/registry.py
kxtran: ✅

zsblevins (approved):

PR #267, opened 2026-08-28
(Chore) - Refactor Device Client

susanhooks:
## Description

Refactor of device client for maintainability. Fixed small issues noted in code rabbit review when refactoring the client. 

## Validation

- [x] Standard CI passes.

```text
/ok to test <sha>
```

```text
/kind test
```

only while debugging.

Kind integration completed with **success** for [`6052471`](https://github.com/dsx-ai-factory/nv-config-manager/commit/60524715d067fefcdd97fc4c89cae03f2e68f4f1) in [this workflow run](https://github.com/dsx-ai-factory/nv-config-manager/actions/runs/33779315178).

## Checklist

## New Features
- Added support for managing Arista, Juniper, Mellanox, Cumulus Linux, and NVOS devices.
- Added configuration comparison, commits, rollbacks, diagnostics, device information, and support-bundle retrieval.
- Added mock device connectivity for local development and testing.
- Added normalized MAC, ARP, and neighbor data handling.

## Bug Fixes
- Improved authentication fallback, secret redaction, error reporting, and connection cleanup.
- Improved handling of malformed network data and failed configuration responses.
- Prevented duplicate ARP relationships and preserved valid data after malformed entries.


src/tests/temporal/clients/device/__init__.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
kxtran: I think that we can remove this file

src/nv_config_manager/temporal/client/device/factory.py
```diff
@@ -0,0 +1,46 @@
...
+    elif device_data.platform == Platform.CUMULUS_LINUX:
+        connection = device_clients.CumulusConnection(device_data.host, site=device_data.site)
+    elif device_data.platform == Platform.NV_OS:
+        connection = device_clients.NVOSConnection(device_data.host, site=device_data.site)
+    elif device_data.platform == Platform.MLNX_OS:
+        connection = device_clients.MellanoxConnection(device_data.host, site=device_data.site)
+    elif device_data.platform == Platform.JUNIPER_JUNOS:
+        connection = device_clients.JuniperConnection(device_data.host, site=device_data.site)
+    else:
+        raise NotImplementedError(f"No handler implemented for platform {device_data.platform}")
```
kxtran: Replaces the `if/elif` platform dispatch with a `match` statement that selects the connection class, so the constructor call appears once. Because `platform` is typed as the `Platform` enum, handling `Platform.UFM` explicitly and
closing with `assert_never` turns exhaustiveness into a mypy error instead of a runtime `NotImplementedError` — same idiom as `workflow_models.intended_config_file`.

src/nv_config_manager/temporal/client/device/factory.py
```diff
@@ -0,0 +1,46 @@
...
+from nv_config_manager.common.config import load_config
+from nv_config_manager.temporal.client.device.base import NetworkConnection
+from nv_config_manager.temporal.common.mixins.device import NetworkDeviceData, Platform
+
+
+def from_device_data(device_data: NetworkDeviceData) -> NetworkConnection:
+    """Return a NetworkConnection for a given device."""
+    # Avoid circular import with device/__init__.py. Resolve classes from the
+    # public package so unittest patches of device.CumulusConnection still apply.
+    from nv_config_manager.temporal.client import device as device_clients
```
kxtran, line 31: Could we rename `from_device_data()` to a package-level `create_connection()`? The concrete implementations could then be imported at module scope in factory.py, with create_connection re-exported from device/__init__.py:
```    
from nv_config_manager.temporal.client.device import create_connection
    with create_connection(device_data) as connection:
        ...
```

This would let us remove `NetworkConnection.from_device_data()` and its local circular-import workaround from base.py, clarify the dependency direction and factory semantics, and provide a natural test patch point: patch create_connection where the caller imports it.

src/nv_config_manager/temporal/client/device/arista.py
```diff
@@ -0,0 +1,421 @@
...
+    def _diff_eq(self, diff_a: str, diff_b: str) -> bool:
+        # Arista session diffs include the session ID in the output
+        # strip those lines when comparing
+        if diff_b == "":
+            # Previous run succeeded to apply but failed on copy run start
+            return True
+        pattern = re.compile(r"session:/.*?-session-config")
+        return pattern.sub("", diff_a) == pattern.sub("", diff_b)
+
+    def perform_candidate_diff(self, new_configuration: str, partial: bool = False) -> str:
```
kxtran, line 241: partial is dropped here, and in `commit_candidate_config` — both call `_load_candidate_config(new_configuration)` positionally. `rollback clean-config` is therefore issued unconditionally and the parameter is dead on this platform.

It's reachable: `TenantDeployWorkflow` sets partial = True (workflows/deploy.py:609) and activities/deploy.py:100/:124 pass it through, so an Arista tenant deploy applies the fragment as the device's entire config — and the approval diff shows a full replacement, so the gate won't catch it.

src/nv_config_manager/temporal/client/device/arista.py
```diff
@@ -0,0 +1,421 @@
...
+        except Exception as exc:  # pylint: disable=broad-except
+            raise NetworkDeviceException(f"Failed to diff session {self._session_id}") from exc
+        finally:
+            self._abort_preserving_pending()
+
+    def commit_candidate_config(
+        self,
+        new_configuration: str,
+        approved_diff: str,
+        partial: bool = False,
```
kxtran: the `partial` flag is not being used here, can we drop it?

src/nv_config_manager/temporal/client/device/cumulus.py
```diff
@@ -0,0 +1,963 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
kxtran, line 1: Only `JuniperConnection` overrides `close()`; Cumulus and Mellanox inherit the base no-op, so the contextlib.closing wrappers in `activities/deploy.py`, `backup.py`, and `diagnostics.py` do nothing for them. Cumulus's pooled requests.Session (cumulus.py:83) and Mellanox's netmiko channel leak until GC; Mellanox's only cleanup is `__del__` (mellanox.py:87), which isn't guaranteed at shutdown. It's worth to adding to both of them

src/nv_config_manager/temporal/client/device/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    ConfigSyntaxException,
+    DiffChangedException,
+    DiffValidationError,
+    InvalidConfigException,
+    NetworkDeviceException,
+)
+from nv_config_manager.temporal.client.device.juniper import (
+    JuniperConnection,
+    _junos_list,
+    _junos_string,
```
kxtran: nit: remove `_junos_list` and `_junos_string` from `__all__ ` have test_juniper.py:39 import them `from nv_config_manager.temporal.client.device.juniper import _junos_list`. This is about not advertising private helpers on the public surface.

src/nv_config_manager/temporal/client/device/models.py
```diff
@@ -0,0 +1,246 @@
...
+            self.mac_to_ip[mac_std] = []
+        self.ip_to_mac[ip_std].append(mac_std)
+        self.mac_to_ip[mac_std].append(ip_std)
+
+    def _add_interface_mac_mapping(self, interface: str, mac_std: str) -> None:
+        """Add interface to MAC mapping."""
+        if interface not in self.interface_to_mac:
+            self.interface_to_mac[interface] = []
+        if mac_std not in self.interface_to_mac[interface]:
+            self.interface_to_mac[interface].append(mac_std)
```
kxtran: These two helpers are private, yet `from_nvue` below re-implements them inline and juniper.py's `_arp_table_from_junos` reaches across the module boundary to call them. Three implementations of one operation.

Collapsing to one public method. Module level:

```
def _append_unique(index: dict[str, list[str]], key: str, value: str) -> None:
    values = index.setdefault(key, [])
    if value not in values:
        values.append(value)
```

and on DeviceArpTable, replacing these helpers:

```
def add_entry(self, ip: str, mac: str, interface: str) -> None:
    """Record one ARP entry, normalising the IP and MAC.
    Raises:
        ValueError / netaddr.core.AddrFormatError: on a malformed IP or MAC.
    """
    ip_std = str(ipaddress.ip_address(ip))
    mac_std = str(netaddr.EUI(mac))
    _append_unique(self.ip_to_mac, ip_std, mac_std)
    _append_unique(self.mac_to_ip, mac_std, ip_std)
    _append_unique(self.interface_to_mac, interface, mac_std)
```
Normalization moves inside (all three callers do it identically) and stays raising so each caller keeps its current handling. The eapi path loops `add_entry` over its comma-split interfaces, `from_nvue` drops its inline block, Juniper stops touching privates.

src/nv_config_manager/temporal/client/device/arista.py
susanhooks: ok added a fix for this

src/nv_config_manager/temporal/client/device/cumulus.py
susanhooks: added

src/nv_config_manager/temporal/client/device/__init__.py
susanhooks: moved and imported

src/nv_config_manager/temporal/client/device/factory.py
susanhooks: Updated this to match the new pattern.

src/nv_config_manager/temporal/client/device/factory.py
susanhooks: I agree this is worth doing, but would prefer to do in a follow on PR to keep this refactor focused.

src/nv_config_manager/temporal/client/device/models.py
susanhooks: deduped and normalized

src/nv_config_manager/temporal/client/device/mellanox.py
```diff
@@ -0,0 +1,168 @@
...
+        self.client.config_mode()
+        output = self.client.send_command(command, expect_string=r"#", read_timeout=timeout)
+        return str(output)
+
+    def close(self) -> None:
+        """Disconnect the netmiko session if it is open."""
+        client = getattr(self, "client", None)
+        if client is None:
+            return
+        self.client = None
```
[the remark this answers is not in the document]
susanhooks: comment above on write memory being only config mode is not accurate

ryanheffernan:
/ok to test [d6626af](https://github.com/dsx-ai-factory/nv-config-manager/pull/267/commits/d6626af953c25474b3820165e6a12c7f17aa80d0)

src/nv_config_manager/temporal/client/device/cumulus.py
```diff
@@ -0,0 +1,980 @@
...
+
+            if link_states[interface]:
+                neighbor_data = lldp_data.get(interface)
+                if neighbor_data:
+                    neighbors[interface] = neighbor_data
+            # Disabling due to performance issues
+            # else:
+            #     info = self.get_ts_info(interface)
+            #     if info.get("link") and info["link"].get("troubleshooting-info"):
+            #         ts_info[interface] = info["link"]["troubleshooting-info"]
```
kxtran: can we remove the commented code?

kxtran (dismissed):
🎉

kxtran (approved):

PR #268, opened 2026-09-01
Move installable packages to package directories

zsblevins:
## Description

Restructure repository layout to clearly separate separately installable packages and plugins.

## Validation

- [ ] Standard CI passes.

```text
/ok to test <sha>
```

```text
/kind test
```

only while debugging.

Kind integration completed with **success** for [`8072eb2`](https://github.com/dsx-ai-factory/nv-config-manager/commit/8072eb2cad151ed2c958838af4e4ffac8610c3ff) in [this workflow run](https://github.com/dsx-ai-factory/nv-config-manager/actions/runs/33554050174).

## Checklist

- [ ] Commits are signed off for DCO compliance.


.gitlab/ci/test.yml
```diff
@@ -21,6 +21,9 @@ test-mr-py:
       changes:
         - "src/**/*"
         - "components/nautobot/**/*"
+        - "packages/dcim/**/*"
+        - "packages/workflows/**/*"
```
kxtran, line 25: could we do "packages/**/*` so it will pick up the templates changes too?

.github/workflows/public-ci.yml
```diff
@@ -40,10 +40,10 @@ jobs:
...
         run: make sort-check
       - run: uv sync --frozen
-      - run: uv run ruff format --check src/ components/network-templates/src components/network-templates/tests packages
-      - run: uv run ruff check src/ components/network-templates/src components/network-templates/tests packages
-      - run: uv run ty check src/nv_config_manager/ packages
-      - run: uv run mypy src/nv_config_manager packages --no-incremental
+      - run: uv run ruff format --check src/ packages plugins/dcim/nautobot-2x
+      - run: uv run ruff check src/ packages plugins/dcim/nautobot-2x
+      - run: uv run ty check src/nv_config_manager/ packages/workflows
+      - run: uv run mypy src/nv_config_manager packages/workflows --no-incremental
```
kxtran: could we also do `packages` here for including templates and dcim?

kxtran (dismissed):

.gitlab/ci/test.yml
zsblevins: Templates already trigger the dedicated lint-mr-network-templates and test-mr-network-templates jobs through packages/templates. Keeping this main Python test trigger scoped to packages/dcim and packages/workflows avoids rerunning unrelated root/DCIM tests for template-only changes.

.github/workflows/public-ci.yml
zsblevins: Before this move, packages contained only workflows, so using all of packages here would newly type-check templates and DCIM rather than preserve the old scope. Templates retain their dedicated Ruff/test coverage; I added packages/dcim/src explicitly because it is clean under both ty and mypy, while keeping the pre-existing templates type debt out of this layout-only PR.

kxtran: that's what I was assuming so I wanted to ask 🙂

kxtran (approved):

PR #277, opened 2026-09-03
feat(helm): decouple Temporal shards from history replicas

spidercensus:
## Description

Decouple the Temporal database shard count from the History service replica count. This allows a cluster to use more persisted History shards without requiring the same number of History pods, because each History pod can serve multiple shards.

This change:

- adds `temporal.numHistoryShards` to the Helm chart;
- preserves existing behavior by deriving shards from `temporal.services.history.replicas` when the new value is omitted;
- exposes History replicas and shards independently through installer YAML and the Workflows TUI;
- omits installer overrides by default so existing install configurations and size profiles remain unchanged;
- documents that legacy deployments derive shards from History replicas, requires pinning the persisted shard count before replica scaling, and directs shard-count changes to an approved migration;
- adds focused schema, installer-values, TUI, and Helm rendering coverage.

## Validation

- `cd installer && uv run pytest -q` — 529 passed.
- `cd installer && uv run pytest tests/test_schema.py tests/test_helm_values.py tests/test_tui.py tests/test_conditional_fields.py -q` — 208 passed after rebasing onto the updated branch.
- `cd installer && uv run ruff check src tests`
- `make docs-lint`
- `helm lint deploy/helm --values deploy/helm/values-ci.yaml`
- `helm template test deploy/helm --values deploy/helm/values-ci.yaml`
- `bash deploy/helm/tests/temporal-history-shards.sh`
- Rendered installer-generated values through Helm and verified 16 History replicas with 128 shards.

- [ ] Standard CI passes.

```text
/ok to test <sha>
```

```text
/kind test
```

only while debugging.

Kind integration completed with **success** for [`6c62f3f`](https://github.com/dsx-ai-factory/nv-config-manager/commit/6c62f3fc0e327087521d79f59d5f9346a67eabb6) in [this workflow run](https://github.com/dsx-ai-factory/nv-config-manager/actions/runs/34627735352).

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added configuration options for Temporal history replicas and shard counts in the installer.
  - Installer settings are available for managed Temporal deployments and validate supported values.
  - Helm now supports explicit Temporal shard-count configuration and reports mismatches with initialized clusters.

- **Documentation**
  - Clarified that Temporal shard counts are established during initialization, and changing them later requires migration.

- **Tests**
  - Added coverage for Temporal sizing, configuration validation, rendering, and shard-count warnings.


ryanheffernan:
this looks good to me, however we should probably make a documentation update to make it clear that the database needs to be migrated to a sharded DB before any shard values changes are made, otherwise the deployment will fail to come up. perhaps some example temporal cli commands for how to perform the re-sharding process would help

spidercensus:
Addressed in 6852b9f. The deployment guide now distinguishes scalable History replicas from the persisted shard count, shows an independent 128-shard/16-replica configuration, and warns that changing the Helm value, editing cluster metadata, or restoring the existing database at a new count is unsafe. Temporal has no in-place re-sharding CLI command; preserving histories requires a separately initialized cluster with independent persistence and an approved multi-cluster replication/handover runbook.

deploy/helm/templates/_temporal-configmap-data.tpl
```diff
@@ -33,7 +33,7 @@ config_template.yaml: |-
     level: DEBUG
 
   persistence:
-    numHistoryShards: {{ .Values.temporal.services.history.replicas }}
+    numHistoryShards: {{ default .Values.temporal.services.history.replicas .Values.temporal.numHistoryShards }}
```
zsblevins: Temporal doesn't actually honor this setting, if the chart is used as is, it'll still get created with 4 shards regardless of this setting: https://community.temporal.io/t/setting-numhistoryshards/14349

We need to adjust the chart where DB creation actually runs to do the proper sharding on initial bringup.

spidercensus: Ack.  I am updating the values with the documentation of that for now.  This is still a draft.  I'll work on the automated migration.

spidercensus: The linked Temporal community thread appears to reference an issue that was seen with a user using the "autosetup" Temporal image, which we are not. I'm testing to verify that the history shards setting in the temporal server config map is respected on initial setup.

spidercensus: I've tested this out pretty extensively.
- On a new deployment, the num history shards parameter is respected and the database is brought online with as many shards as requested.  It's also decoupled from the worker count, which is a feature we implemented in this PR.
- On an existing deployment, if I change the num history shards parameter and run Helm upgrade, the setting is ignored -- i.e. the temporal server ConfigMap is updated to contain the configured value, but the database still shows the same number of shards.
To verify config map value: 
```
➜  ~ kubectl get configmap nv-config-manager-temporal-config -n "$NS" \
  -o jsonpath='{.data.config_template\.yaml}' |
  grep -F "numHistoryShards: ${TARGET_SHARDS}"
```

To verify database shard count, run this in CNPG on the main temporal db:
```
SELECT COUNT(*) AS history_shard_count FROM shards;
```

ryanheffernan (dismissed):

